PR #4, opened 2020-01-13
feat(platforms): init

divyenduz:
- [x] Netlify via CLI
- [x] Heroku


platforms/heroku/index.js
```diff
@@ -0,0 +1,48 @@
...
+    },
+  })
+
+  const users = await photon.users.findOne({
+    where: {
+      id: createUser.id,
+    },
+  })
+
+  const deleteManyUsers = await photon.users.deleteMany()
```
steebchen: afaik this should be `.deleteMany({})`

platforms/heroku/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+  return res.send(
+    JSON.stringify({
```
steebchen: pretty sure you can just do `res.json({ ... }` without `JSON.stringify` :)

platforms/heroku/index.js
divyenduz: types give a hiccup at times, but this is "js" and runtime is happy with this.

platforms/heroku/index.js
divyenduz: Not super sure, will give it a go tomorrow

steebchen:
just FYI, the frameworks error should disappear after upgrading prisma2 (so it'll be fixed automatically when you merge master into this branch or just merge into master)

PR #7, opened 2020-01-15
feat(platforms): use project-specific DBs

steebchen:
This makes sure #6 will work without conflicts by giving every project its own DB.


steebchen:
do not merge yet!

PR #14, opened 2020-01-22
netlify + manually generate

divyenduz:
For some reason, postinstall hook from facade doesn't always work in netlify. This command entry ensures that Prisma client exists.


divyenduz:
Nope! It is just flaky in the CI.

PR #16, opened 2020-01-24
Netlify + ZISI flag

divyenduz:
WIP: checking if netlify's current CI coupled with the ZISI flag works with Prisma.


divyenduz:
Disabled the actual test for the platform as it doesn't work (making the test green). Merging this because I want to keep: 

1. The infra to test this further once I hear back from netlify support (maintaining a long living branch would be a pain)
2. This has unrelated changed that I needed to do. Master was broken.

PR #26, opened 2020-01-31
feat(gcp-functions): use a new function for each test run

steebchen:


steebchen:
note: CSB is red for unrelated reasons; merging anyway since it's not affected by GCP

PR #32, opened 2020-02-04
feat(bundlers): init parcel

divyenduz:
Related https://github.com/parcel-bundler/parcel/issues/355


divyenduz:
It failed on netlify, which is probably intermittent. Merging this shortlt.

PR #153, opened 2020-02-25
docs(readme): fix filename, remove duplicated information, adapt list

janpio:


steebchen:
Note: libraries are currently not included in the readme, since they'll be moved to prisma-examples soon.

PR #242, opened 2020-03-27
fix: exit code, use process.exit 1

divyenduz:


divyenduz:
Most of the other code has `process.exit(1)`, I just followed that convention. Let's merge this then and create an isolated repro for the panic? wdyt?

If you free strongly about changing it to `throw e`, I can quickly make a commit and replace it across the code base.

steebchen:
Ok, maybe I remembered incorrectly... let's merge this for now, not that important.

PR #255, opened 2020-03-29
fix(all): Adapt schema to new relations

janpio:
alpha 978 brought a new relation format for the schemas, which broke most e2e tests. This PR updates the project to apply that.


steebchen:
Thanks ❤️

PR #329, opened 2020-04-16
fix(netlify): use the new url

divyenduz:
- maybe we should use `curl -L` (follow redirects flag)


divyenduz:
Fixes https://github.com/prisma/prisma2-e2e-tests/issues/328

PR #374, opened 2020-05-14
fix(netlify): for netlify git deployments

divyenduz:
remove node_modules before `yarn install`


divyenduz:
image

Why does this PR expect platform `zeit-now`?

PR #389, opened 2020-06-04
feat: Update nexus package name

jasonkuhrt:


frameworks/nexus/package.json
```diff
@@ -10,15 +10,16 @@
...
+    "@prisma/client": "^2.0.0-beta.4",
     "graphql": "14.6.0",
     "graphql-yoga": "1.18.3",
-    "nexus": "0.12.0",
-    "nexus-prisma": "0.11.1"
+    "nexus-prisma": "^0.14.0"
   },
   "devDependencies": {
-    "@prisma/cli": "2.0.0-alpha.1311",
+    "@prisma/cli": "2.0.0-beta.4",
```
janpio, line 20: This won't work as the e2e test main functionality is that they update this version (and client above) every time we release a new version of that. So renovate would just create a new PR again. 

Or is that what we want as this combination here should work, and if the update does not we know that nexus schema is not ready yet?

jasonkuhrt: I'm not familiar with alphas. I know beta 4 is what the current prisma plugin stable versions are suing. Sounds like this system is going by its own strategy though so no interest in going against it.

jasonkuhrt: Actually pardon me, technically `nexus-prisma` needs beta 4, anything else is outside what it obliges itself to work against.

jasonkuhrt: None the less I've reverted my changes.

frameworks/nexus/package.json
```diff
@@ -10,15 +10,16 @@
...
     "graphql": "14.6.0",
     "graphql-yoga": "1.18.3",
-    "nexus": "0.12.0",
-    "nexus-prisma": "0.11.1"
+    "nexus-prisma": "^0.14.0"
   },
   "devDependencies": {
     "@prisma/cli": "2.0.0-alpha.1311",
     "@types/node": "12.12.37",
+    "@types/react": "^16.9.35",
```
jasonkuhrt, line 22: This is required becuase of https://github.com/graphql-nexus/nexus-schema-plugin-prisma/issues/675

PR #481, opened 2020-06-29
feat: heroku and pgbouncer buildpack

divyenduz:
Closes https://github.com/prisma/e2e-tests/issues/469

Notes: 

1. This belongs in `platforms` as it is only possible to use a build pack in conjunction with a Heroku app.
1. Adding the pgbouncer buildpack automatically configures pgbouncer against the database in `DATABASE_URL` environment variable and alters adds `DATABASE_URL_PGBOUNCER` at runtime, that can be used.
1. We did not have to use our pgbouncer connection string parameter for this.
1. In theory, this should work in any postgres database but when I used a connection string from AWS, it failed to parse it properly (username, password fields were just missing). I continued with a Heroku postgres database for now. 
1. ~Tests are not running right now because GH is facing issues: https://www.githubstatus.com/. This blocks the merge of this PR.~ Tests ran successfully, I will merge this shortly.


janpio:
> This belongs in `platforms` as it is only possible to use a build pack in conjunction with a Heroku app.

How is that connected? 


> We did not have to use our pgbouncer connection string parameter for this.

What does that mean?

> In theory, this should work in any postgres database but when I used a connection string from AWS, it failed to parse it properly (username, password fields were just missing). I continued with a Heroku postgres database for now.

So Heroku PgBouncer with AWS database?

divyenduz:
> How is that connected?

It is no longer "purely database", `platforms` suits it the best

> What does that mean?

I did not have to use `pgbouncer=true` in the database connection string

> So Heroku PgBouncer with AWS database?

Heroku PgBouncer with AWS database didn't work because the user name was not parsed. I ended up using Heroku postgres addon.

PR #491, opened 2020-06-30
chore: move platforms/heroku-pgbouncer-buildpack to databases/heroku-pgbouncer-buildpack

divyenduz:


janpio:
When moving checks, we also need to upgrade the required checks of the `dev` branch.

divyenduz:
> When moving checks, we also need to upgrade the required checks of the dev branch.

That was done right after the merge.

PR #498, opened 2020-06-30
chore: upgrade nexus-prisma

jasonkuhrt:


jasonkuhrt:
@janpio would it help if I did in this PR now:

```json
{
  "resolutions": {
    "@prisma/client": "2.2.0-dev.27",
    "@prisma/cli": "2.2.0-dev.27"
  }
}
```

janpio:
Nope, as that would just lead to us having a project that is not updated automatically as it should be. 
Only when the issue https://github.com/prisma/e2e-tests/issues/499 is fixed, will the update process take that into account.

PR #509, opened 2020-07-01
feat(databases/bedrock): Add Bedrock database

janpio:
TODO

- [ ] README
- [ ] Binaries for newer Ubuntu
- [ ] Migrate should also work instead of having to commit the database (although that is kinda out of scope for these tests)


.github/workflows/test.yaml
```diff
@@ -300,7 +300,7 @@ jobs:
           LIBRARY_TYPE_GRAPHQL_PG_URL: ${{ secrets.LIBRARY_TYPE_GRAPHQL_PG_URL }}
 
   databases:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-16.04
```
janpio, line 561: no prebuilt bedrock binaries for 18 or 20 right now published

PR #527, opened 2020-07-07
fix(frameworks/nestjs): adapt deps & fix tsconfig

steebchen:


steebchen:
Closing because nests depends on apollo-server-express but it's impossible to upgrade.

PR #528, opened 2020-07-07
fix(frameworks/nextjs): update nextjs

steebchen:
- upgrade nextjs
- use unstable_revalidate
- make getStaticProps stable


steebchen:
Merging with admin rights because there constant conflicts with prisma reps, tests are passing.

PR #560, opened 2020-07-15
feat: customize emoji on the last attempt

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/505

How: 

1. We want to know if the current step execution is the last attempt from the `retry` action
2. To add this feature to the `retry` action I tried using `setOutput` and `exportVariable` from `@actions/core` but a step can only set outputs/env vars for upcoming steps not the step itself
3. But we can use the filesystem to persist the state of an action, provided we scope the filename to be unique across all steps, this is implemented via https://github.com/nick-invision/retry/pull/8
4. We are using the branch directly from the `retry` action PR in step 3, so a merge is not required
5. Using `filename_last_attempt` input, we provide the `retry` action with our desired filename. `test-project.sh` has all the information to re-create this filename
6. We read the current state of a step and change the response based on that information

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -10,6 +10,16 @@ set +u
 matrix=$3
 set -u
 
+if [ -z "$matrix" ]; then
+	FILENAME_LAST_ATTEMPT="$dir-$project"
+else
+	FILENAME_LAST_ATTEMPT="$dir-$project-$matrix"
+fi
+echo "FILENAME_LAST_ATTEMPT $FILENAME_LAST_ATTEMPT"
+
+LAST_ATTEMPT=$(touch "$FILENAME_LAST_ATTEMPT" && cat "$FILENAME_LAST_ATTEMPT")
+echo "LAST_ATTEMPT $LAST_ATTEMPT"
+
 sh .github/scripts/print-version.sh "$dir/$project/package.json"
 
 cd .github/slack/
@@ -88,7 +98,7 @@ echo "$dir/$project done"
 
 cd "$root"
 
-if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-dev" ] || [ "$GITHUB_REF" = "refs/heads/latest" ]; then
+if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-dev" ] || [ "$GITHUB_REF" = "refs/heads/latest" ] || [ "$GITHUB_REF" = "refs/heads/feat_last_attempt" ]; then
 	(cd .github/slack/ && yarn install --silent)
 
 	branch="${GITHUB_REF##*/}"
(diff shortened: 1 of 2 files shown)


.github/scripts/test-project.sh
```diff
@@ -88,7 +98,7 @@ echo "$dir/$project done"
 
 cd "$root"
 
-if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-dev" ] || [ "$GITHUB_REF" = "refs/heads/latest" ]; then
+if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-dev" ] || [ "$GITHUB_REF" = "refs/heads/latest" ] || [ "$GITHUB_REF" = "refs/heads/feat_last_attempt" ]; then
```
divyenduz, line 101: TODO: remove this new ref before the merge

.github/scripts/test-project.sh
```diff
@@ -107,12 +117,18 @@ if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-d
 	fi
 
 	echo "notifying slack channel"
-	node .github/slack/notify.js "prisma@$version: ${emoji} $workflow_link ran (via $commit_link)"
+	node .github/slack/notify.js "TEST: prisma@$version: ${emoji} $workflow_link ran (via $commit_link)"
```
divyenduz, line 120: TODO: remove 'TEST:" before merge

.github/scripts/test-project.sh
```diff
@@ -107,12 +117,18 @@ if [ "$GITHUB_REF" = "refs/heads/dev" ] || [ "$GITHUB_REF" = "refs/heads/patch-d
...
 		export webhook="$SLACK_WEBHOOK_URL_FAILING"
 		echo "notifying failing slack channel"
-		node .github/slack/notify.js "prisma@$version: :x: $workflow_link failed (via $commit_link)"
+
+		emoji=":question:"
+		if [ $LAST_ATTEMPT = 'true' ]; then
+			emoji=":x:"
+		fi
+
+		node .github/slack/notify.js "TEST: prisma@$version: ${emoji} $workflow_link failed (via $commit_link)"
```
divyenduz, line 131: TODO: remove 'TEST:" before merge

.github/workflows/test.yaml
```diff
@@ -47,6 +51,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v1
+        with:
+          ref: feat_last_attempt
```
divyenduz, line 55: Remove all instances of this ref before the merge, so we are checking out master when in master.

divyenduz:
This will probably be superseded by https://github.com/prisma/e2e-tests/pull/561 which is much simpler but has one limitation over this approach. 

In this approach, we can granularly control the last failure message, so max attempts 3 can be represented as

In this stateful approach: ❓,❓,❌ 
In the simpler approach: ❓,❓,❓,❌

divyenduz:
Closed in favor of https://github.com/prisma/e2e-tests/pull/561

PR #599, opened 2020-07-23
fix(library/nexus-schema): drop graphql-yoga for express

jasonkuhrt:


libraries/nexus-schema/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
+    "esModuleInterop": true,
```
jasonkuhrt, line 3: for express

libraries/nexus-schema/src/server.ts
```diff
@@ -1,9 +1,19 @@
...
+  '/graphql',
+  graphqlHTTP({
+    schema,
+    graphiql: true,
+  }),
 )
+
+app.listen(4000, () => {
+  console.log(
+    `🚀 Server ready at: http://localhost:4000/graphql\n⭐️ See sample queries: http://pris.ly/e/ts/graphql#3-using-the-graphql-api`,
```
janpio:
```suggestion
    `🚀 Server ready at: http://localhost:4000/graphql`,
```

jasonkuhrt:
Able to reproduce locally https://github.com/prisma/e2e-tests/pull/599/checks?check_run_id=902653053#step:4:107

Don't understand why it is happening.

Able to reproduce locally https://github.com/prisma/e2e-tests/pull/599/checks?check_run_id=902653053#step:4:115

Don't understand why it is happening.

jasonkuhrt:
@janpio should be ready to go now

libraries/nexus-schema/src/server.ts
```diff
@@ -1,15 +1,17 @@
...
 import { schema } from './schema'
 
 const app = express()
 
 app.use(
   '/',
-  graphqlHTTP({
+  graphqlHTTP((req) => ({
     schema,
+    context: createContext(),
```
janpio, line 12: Any idea why this is needed @jasonkuhrt? The examples on https://github.com/graphql/express-graphql#simple-setup don't use it.

janpio (approved):

libraries/nexus-schema/src/server.ts
jasonkuhrt: That's the way resolver context gets created for every request. If a developer does not provide it, then express-graphql defaults to making the resolver context be the request object.

jasonkuhrt: https://github.com/graphql/express-graphql/issues/187

jasonkuhrt:
@janpio is the policy here for devs to merge their own PRs after approval?

libraries/nexus-schema/src/server.ts
janpio: Why does the default not work for our project?

janpio:
We do not have clear guideline for this - so I will just merge this when the test have run.

libraries/nexus-schema/src/server.ts
jasonkuhrt: Because we need to expose prisma client on the resolver context.

I don't think we need a context function though. That's just how it was setup before.

PR #654, opened 2020-08-04
chore: reconnect after first query only

divyenduz:
Clean up of https://github.com/prisma/e2e-tests/pull/651

### DIFF
--- a/databases/digitalocean-pgbouncer/index.js
+++ b/databases/digitalocean-pgbouncer/index.js
@@ -21,24 +21,26 @@ const clientWithQueryStringParam = new PrismaClient({
 })
 
 async function clientWithoutQueryStringParamCall() {
-  await client.disconnect()
-  await client.connect()
   const data = await client.user.findMany()
   return data
 }
 
 async function clientWithQueryStringParamCall() {
-  await clientWithQueryStringParam.disconnect()
-  await clientWithQueryStringParam.connect()
   const data = await clientWithQueryStringParam.user.findMany()
   return data
 }
 
 async function main() {
-  const data1 = await client.user.findMany()
-  console.log(data1)
-  const data2 = await client.user.findMany()
-  console.log(data2)
+  try {
+    const data1 = await client.user.findMany()
+    console.log(data1)
+    await client.disconnect()
+    await client.connect()
+    const data2 = await client.user.findMany()
+    console.log(data2)
+  } catch (e) {
+    console.log(e)
+  }
 
   const data3 = await clientWithQueryStringParam.user.findMany()
   console.log(data3)
--- a/databases/digitalocean-pgbouncer/index.js
+++ b/databases/digitalocean-pgbouncer/index.js
@@ -34,8 +34,10 @@ async function main() {
   try {
     const data1 = await client.user.findMany()
     console.log(data1)
+
     await client.disconnect()
     await client.connect()
+    
     const data2 = await client.user.findMany()
     console.log(data2)
   } catch (e) {
--- a/databases/digitalocean-pgbouncer/index.js
+++ b/databases/digitalocean-pgbouncer/index.js
@@ -35,9 +35,15 @@ async function main() {
     const data1 = await client.user.findMany()
     console.log(data1)
 
+    /*
+    * Query engine instance names prepared statements serially s0, s1 and so on. Without the pgbouncer=true flag, 
+    * prepared statements are not cleaned up in PgBouncer. By doing disconnect/reconnect, we get a 
+    * new insatnce of query engine that starts again at s0. And we expect the next client call to throw
+    * prepared statement s0 already exists
+    */
     await client.disconnect()
     await client.connect()
-    
+
     const data2 = await client.user.findMany()
     console.log(data2)
   } catch (e) {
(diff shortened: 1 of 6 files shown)


databases/digitalocean-pgbouncer/index.js
```diff
@@ -21,24 +21,26 @@ const clientWithQueryStringParam = new PrismaClient({
...
 async function main() {
-  const data1 = await client.user.findMany()
-  console.log(data1)
-  const data2 = await client.user.findMany()
-  console.log(data2)
+  try {
+    const data1 = await client.user.findMany()
+    console.log(data1)
+    await client.disconnect()
+    await client.connect()
```
janpio, line 45:
```suggestion

    await client.disconnect()
    await client.connect()
    
```

databases/digitalocean-pgbouncer/index.test.js
```diff
@@ -15,7 +15,10 @@ describe('should test Prisma client and PgBouncer', () => {
   it('should fail with default do pgbouncer without the pgbouncer query string param', async () => {
     try {
       await clientWithoutQueryStringParamCall()
+      await client.disconnect()
+      await client.connect()
```
janpio, line 26:
```suggestion

      await client.disconnect()
      await client.connect()
      
```

databases/docker-pgbouncer/index.test.js
```diff
@@ -15,7 +15,10 @@ describe('should test Prisma client and PgBouncer', () => {
   it('should fail with docker pgbouncer without the pgbouncer query string param', async () => {
     try {
       await clientWithoutQueryStringParamCall()
+      await client.disconnect()
+      await client.connect()
```
janpio, line 26:
```suggestion

      await client.disconnect()
      await client.connect()

```

databases/heroku-pgbouncer/index.js
```diff
@@ -21,24 +21,26 @@ const clientWithQueryStringParam = new PrismaClient({
...
 async function main() {
-  const data1 = await client.user.findMany()
-  console.log({ data1 })
-  const data2 = await client.user.findMany()
-  console.log({ data2 })
+  try {
+    const data1 = await client.user.findMany()
+    console.log({ data1 })
+    await client.disconnect()
+    await client.connect()
```
janpio, line 45:
```suggestion

    await client.disconnect()
    await client.connect()
    
```

databases/heroku-pgbouncer/index.test.js
```diff
@@ -15,7 +15,10 @@ describe('should test Prisma client and PgBouncer', () => {
   it('should fail with default heroku pgbouncer without the pgbouncer query string param', async () => {
     try {
       await clientWithoutQueryStringParamCall()
+      await client.disconnect()
+      await client.connect()
```
janpio, line 26:
```suggestion

      await client.disconnect()
      await client.connect()
      
```

janpio (commented):
All these disconnect+connect should probably get a comment what they do - either in code or in the README that explains all these tests.

janpio (commented):
Do we maybe have a way to unify these into 1 folder and just call it with different parameters to use the different connection strings? Would cut down on code duplication a lot and allow us to make sure that all pgbouncer tests use the exact same tests etc.

divyenduz:
> Do we maybe have a way to unify these into 1 folder and just call it with different parameters to use the different connection strings? Would cut down on code duplication a lot and allow us to make sure that all pgbouncer tests use the exact same tests etc.

Would make it difficult to trace if one of the platforms has an issue in future due to any change. Also would make the tests difficult to read as we add more platforms.

divyenduz:
> All these disconnect+connect should probably get a comment what they do - either in code or in the README that explains all these tests.

Added inline comment.

janpio:
> Would make it difficult to trace if one of the platforms has an issue in future due to any change. Also would make the tests difficult to read as we add more platforms.

These would still need to be tested individually - just instead of duplicating the code x times we would have 1 folder that runs multiple tests.

databases/digitalocean-pgbouncer/index.js
```diff
@@ -35,9 +35,15 @@ async function main() {
     const data1 = await client.user.findMany()
     console.log(data1)
 
+    /*
+    * Query engine instance names prepared statements serially s0, s1 and so on. Without the pgbouncer=true flag, 
+    * prepared statements are not cleaned up in PgBouncer. By doing disconnect/reconnect, we get a 
+    * new insatnce of query engine that starts again at s0. And we expect the next client call to throw
+    * prepared statement s0 already exists
```
janpio:
```suggestion
    * Query engine instance names prepared statements serially s0, s1 and so on. Without the `pgbouncer=true` flag, 
    * prepared statements are not cleaned up in PgBouncer. By doing disconnect/reconnect, we get a 
    * new instance of query engine that starts again at s0. And we expect the next client call to throw
    * "prepared statement s0 already exists"
```

divyenduz:
> These would still need to be tested individually - just instead of duplicating the code x times we would have 1 folder that runs multiple tests.

I still wouldn't prefer that. To detect a specific failing platform you would still have to read and understand multiple "unrelated" tests vs check/no-check for a specific platform.

Besides, code duplication is all across the E2E repo and it isn't a lot of code.

PR #664, opened 2020-08-05
chore: rename now.json to vercel.json

styfle:
- Rename `now.json` to `vercel.json`
- Change `@now/node` to `@vercel/node`


platforms/vercel-node-builder/vercel.json
```diff
@@ -8,5 +8,5 @@
       "VERCEL_NODE_BUILDER_PG_URL": "@vercel_node_builder_pg_url"
     }
   },
-  "builds": [{ "src": "*.js", "use": "@now/node" }]
+  "rewrites": [{ "source": "(.*)", "destination": "/api/app" }]
```
divyenduz: @styfle In this test we wanted to test `@now/node` builder specifically, there was an issue with the builder while the `api` folder approach to create an API still worked. 

Is there a vercel equivalent of `@now/node`? or is this change equivalent?

styfle: I see. It should be using the `@vercel/node` package instead.

divyenduz: Thanks, and probably this would need to be reverted too, right?

`platforms/vercel-node-builder/index.js → platforms/vercel-node-builder/api/app.js`

styfle: Yep, done 👍

divyenduz:
Thank you for the PR @styfle 🙌

PR #730, opened 2020-08-25
docs(platforms): separate serverless and other platforms

janpio:
Functions/Serverless vs. others.

All correct?

### DIFF
--- a/platforms/README.md
+++ b/platforms/README.md
@@ -1,0 +1,19 @@
+# Platforms
+
+## Serverless
+
+- azure-functions-linux
+- azure-functions-windows
+- firebase-functions
+- gcp-functions
+- lambda
+- netlify-beta-ci
+- netlify-current-cli
+- serverless-lambda
+- vercel-api
+- vercel-node-builder
+
+## Others
+
+- codesandbox
+- heroku


steebchen:
I guess it depends on the definition of server less. I don't really see the need for it, but the differentiation seems logical to me.

steebchen (approved):

steebchen:
(not sure why the semantic bot doesn't pick up the semantic commit change – it should)

divyenduz (approved):

divyenduz:
> (not sure why the semantic bot doesn't pick up the semantic commit change – it should)

I think because the commit message doesn't have a `:`

janpio:
> 
> 
> > (not sure why the semantic bot doesn't pick up the semantic commit change – it should)
> 
> I think because the commit message doesn't have a `:`

It _should_ be happy with either the commit message or the PR title. But will be fixed on merge.

> I guess it depends on the definition of server less. 

Can you elaborate on that? I used this in the way is usually used in the market now to mean function like, stateless code execution.

steebchen:
> I think because the commit message doesn't have a :

Yes, but the (edited) title should have precedence. I was actually quite sure that it would.

steebchen:
> Can you elaborate on that?

Let's not, fine by me to merge.

janpio:
> > I think because the commit message doesn't have a :
> 
> Yes, but the (edited) title should have precedence. I was actually quite sure that it would.

Same. Seems to be a bug: https://github.com/zeke/semantic-pull-requests/issues/112#issue-686265217

PR #756, opened 2020-08-31
feat(google): add postinstall prisma telemetry information

steebchen:


platforms/firebase-functions/functions/package.json
```diff
@@ -7,7 +7,7 @@
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "postinstall": "CI=1 yarn prisma generate"
+    "postinstall": "CI=1 && PRISMA_TELEMETRY_INFORMATION='e2e-tests platform postinstall' && yarn prisma generate"
```
janpio, line 10: Is this run after `npm install` is run in a project that contains Prisma CLI or Client? 
That might unfortunately already be too late if the normal Client `postinstall` hook was triggered :/

steebchen: I would need to double check, but even if this is needed indeed we would need to set an env var somewhere else as well

steebchen: as far as I remember the only which ran is this manual post install hook, but as I said I would need to double check by messing around with gcp functions

janpio: Ok, I would have loved to use Telemetry to actually find that out - but right now the string does not allow me to differentiate between the platforms. Fixed in https://github.com/prisma/e2e-tests/pull/760

janpio: Actually, already strange: https://app.redash.io/prismaio/queries/454870?p_information=e2e-tests%20platform%20postinstall These are all identified as "Github Actions". Is that code only executed in Github actions and not on the deployment host maybe?

janpio: But also: https://app.redash.io/prismaio/queries/398924?p_signature=b8dcfb27-8ef7-4dd3-a7fb-d2b88a4baf9e This indicates that some `npm install` is indeed run before this code can overwrite it again.

steebchen: ah well, the post install hook is also run on GitHub Actions for running the tests itself

janpio: Yep, confirmed again:
https://app.redash.io/prismaio/queries/398924?p_signature=a90f4ab6-137c-4f46-b36d-28edf67dd794
https://app.redash.io/prismaio/queries/398924?p_signature=ebd8816f-e95c-41c1-a531-d84b2a9931b7

So there is already a `postinstall` from Prisma CLient run before you set this ENV var.
And this seems to run on GH Actions instead of Firebase or GCP.

steebchen: ah, good to know

PR #764, opened 2020-09-01
chore: add telemetry info

divyenduz:
Apart from this PR, I also added the following from the UI of respective projects: 

|Project|PRISMA_TELEMETRY_INFORMATION|
|---|---|
| databases/heroku-pgbouncer-buildpack | e2e-tests databases heroku-pgbouncer-buildpack ci |
| databases/heroku-pgbouncer | e2e-tests databases heroku-pgbouncer ci |
| platforms/heroku | e2e-tests platforms heroku ci |
| platforms/netlify-beta-ci | e2e-tests platforms netlify-ci ci |
| platforms/netlify-current-cli | e2e-tests platforms netlify-cli ci |
| platforms/vercel-api | e2e-tests platforms vercel-api ci |
| platforms/vercel-node-builder | e2e-tests platforms vercel-node-builder ci |

### DIFF
--- a/databases/heroku-pgbouncer-buildpack/run.sh
+++ b/databases/heroku-pgbouncer-buildpack/run.sh
@@ -3,6 +3,7 @@
 set -eux
 
 sudo snap install heroku --classic
+export PRISMA_TELEMETRY_INFORMATION='e2e-tests databases heroku-pgbouncer-buildpack build'
 yarn install
 yarn prisma generate
 
--- a/databases/heroku-pgbouncer/run.sh
+++ b/databases/heroku-pgbouncer/run.sh
@@ -2,5 +2,6 @@
 
 set -eu
 
+export PRISMA_TELEMETRY_INFORMATION='e2e-tests databases heroku-pgbouncer build'
 yarn install
 yarn prisma generate
--- a/platforms/heroku/run.sh
+++ b/platforms/heroku/run.sh
@@ -3,6 +3,7 @@
 set -eux
 
 sudo snap install heroku --classic
+export PRISMA_TELEMETRY_INFORMATION='e2e-tests platforms heroku build'
 yarn install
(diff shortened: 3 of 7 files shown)


divyenduz:
`databases/heroku-pgbouncer-buildpack`: build, ci (I had missed adding the CI env var at the first run)

image
image

`databases/heroku-pgbouncer`: build (no ci because we just use the database)

image


`platforms/heroku`: build, ci -> If it runs in both places, one might be a noop

image
image


`platforms/netlify-beta-ci`: ci

image

`platforms/netlify-current-cli`: build

image

`platforms/vercel-api`: build, ci -> If it runs in both places, one might be a noop

image
image

`platforms/vercel-node-builder`: build, ci -> If it runs in both places, one might be a noop

image

image

steebchen (commented):
The env vars in the .sh files directly followed by generate are probably duplicate tagging because they would be already tagged by the action env var, right? I think it would be enough to just tag those who aren't covered by that

divyenduz:
> The env vars in the .sh files directly followed by generate are probably duplicate tagging because they would be already tagged by the action env var, right? I think it would be enough to just tag those who aren't covered by that

This gives us more granularity by marking each project separately vs action env var which is just `e2e-tests test.yaml`. And this var is set before `yarn install`, so it would be duplicated only where `test.yaml` also does `yarn install` in which case we can probably remove one of the `yarn install` commands. 

My rationale of doing it like this without changing any implementation was that we add telemetry info on top of today's reality, might explain some things that we have seen in the CI and then we can optimise by removing duplicates etc. What do you think?

steebchen:
Right, but maybe we still could have done this in a unified way where it's set only once – we get the variables from the script (e.g. in test-project.sh) and then just manually add the env var where it's additionally required (like on a remote npm install).

PR #793, opened 2020-09-08
ci(test-all): link to individual GitHub runs

steebchen:
Instead of linking to the workflow run, we want the actions to link to the exact job. This is just a draft of how we can access the GitHub API; unfortunately it turns out it's not possible to get individual check run IDs yet. Once it does, we may want to merge, but as long as it doesn't, we don't.

image


divyenduz:
Closing this for now

PR #802, opened 2020-09-09
fix(ci): make binaries job clearer

janpio:
Right now one could think we test our normal binaries on 3 OSes. 
In reality we are testing a special `pkg` binary on these OSes.


steebchen (approved):
Smart!

PR #808, opened 2020-09-10
feat(redwood): init

### LINKED ISSUE #681: Redwood on Vercel with Prisma

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/681


steebchen (approved):

platforms/vercel-with-redwood/.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
```
steebchen: we shouldn't need a per-project editorconfig

platforms/vercel-with-redwood/.env.example
```diff
@@ -0,0 +1,2 @@
+# DATABASE_URL=file:./dev.db
```
steebchen, line 1: ??

platforms/vercel-with-redwood/.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
```
steebchen: maybe also remove this? not really important to this repo since it's just one of many

platforms/vercel-with-redwood/.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
steebchen: same

platforms/vercel-with-redwood/.editorconfig
divyenduz: I think this comes from redwood started project. I will remove it, if needed.

platforms/vercel-with-redwood/.env.example
divyenduz: It comes from redwood starter project.

platforms/vercel-with-redwood/.editorconfig
divyenduz: Removed.

.github/workflows/test.yaml
```diff
@@ -332,7 +335,14 @@ jobs:
...
       - uses: actions/checkout@v1
 
+      - name: use node 12
+        if: ${{ matrix.platform != 'firebase-functions' }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
       - name: use node 10
+        if: ${{ matrix.platform == 'firebase-functions' }}
```
janpio, line 345: ?

platforms/vercel-with-redwood/api/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "api",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
+    "@prisma/cli": "2.7.0-dev.43",
+    "@prisma/client": "2.7.0-dev.43",
```
janpio, line 7: This does not make sense at the same time as `resolutions` I think?

janpio (commented):
So much generated code 👍

.github/workflows/test.yaml
steebchen: I discussed this with @divyenduz; we need to use node 12 for redwood so we just upgrade all to node 12, but firebase currently only runs on node 10 so we make an exception for firebase.

platforms/vercel-with-redwood/api/package.json
divyenduz: Removed.

divyenduz: I think that is still required. I [removed it](https://github.com/prisma/e2e-tests/pull/808/commits/3ffefd337e4d5fd0c98b204367feab0347175f15) and the test failed again 

Expected: "2.7.0-dev.43"
Received: "2.6.1"

Bringing it back!

.github/workflows/test.yaml
janpio: A comment in the code would be welcome as this context is impossible to get for other people otherwise.

divyenduz: https://github.com/prisma/e2e-tests/commit/3ff6662176d98c2a4ce7b7275283288be9a9c764

divyenduz: Created a tracking issue and linked it via https://github.com/prisma/e2e-tests/commit/3ff6662176d98c2a4ce7b7275283288be9a9c764

platforms/vercel-with-redwood/api/package.json
steebchen: it's not required: https://github.com/prisma/e2e-tests/blob/dd9d0cbb77cc55ec2f8e5b655c6eb7797f71300e/libraries/nexus-schema/package.json#L12-L31

it may have failed because of something else

divyenduz: It failed because the versions didn't match anymore, I can explore this further at a later time. 

The received version was the one that is bundled with Redwood. 

```
Expected: "2.7.0-dev.43"
Received: "2.6.1"
```

divyenduz: Interestingly, just resolutions don't work (probably because this is a workspace - I read an issue that in workspaces resolutions don't work yet) but just using deps works https://github.com/prisma/e2e-tests/pull/816

PR #822, opened 2020-09-12
ci: identify codesandbox workflow

janpio:


platforms/codesandbox/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "",
   "main": "src/index.js",
   "scripts": {
-    "prestart": "CI=1 yarn prisma migrate up --auto-approve --experimental",
+    "preinstall": "export CI=1 && export PRISMA_TELEMETRY_INFORMATION='e2e-tests platform codesandbox preinstall'",
+    "prestart": "CI=1 && PRISMA_TELEMETRY_INFORMATION=\"e2e-tests platform codesandbox prestart\" && yarn prisma migrate up --auto-approve --experimental",
```
janpio: I don't think this is actually working.

platforms/codesandbox/prisma/.env
```diff
@@ -0,0 +1,2 @@
+CI=1
```
janpio: This somehow also does not work :/

divyenduz: Same, the container did get the correct `.env` file https://codesandbox.io/s/j6txg?file=/prisma/.env

platforms/codesandbox/package.json
divyenduz: The container has the same package.json and that is where `npm install` is run, so it should work https://codesandbox.io/s/j6txg?file=/package.json

platforms/codesandbox/prisma/.env
janpio: No events with `CI=true` for `%codesandbox%`:
https://app.redash.io/prismaio/queries/469381?p_ci=%5B%22false%22%2C%22true%22%5D&p_information=codesandbox

platforms/codesandbox/package.json
janpio: Unfortunately only the value from `.env` is being picked up: 
https://app.redash.io/prismaio/queries/469381?p_ci=%5B%22false%22%2C%22true%22%5D&p_information=codesandbox

janpio: Removed from the file to not create wrong expectations.

janpio: Removed from the file to not create wrong expectations.

platforms/codesandbox/prisma/.env
janpio: Added comment to indicate this does not work.

janpio: https://github.com/prisma/e2e-tests/issues/827

PR #823, opened 2020-09-15
fix(redwood): add -W to yarn install

divyenduz:
For reasons unknown, redwood test started to throw this error in the update script 

```
error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
```

The fix is to add -W to the yarn add/install commands. But since we don't have an easy way to do it specifically to a project in `check-for-update.sh`, I added that generally (which shouldn't cause any issues)


divyenduz:
Superseded by https://github.com/prisma/e2e-tests/pull/824

PR #857, opened 2020-10-12
feat: /rebase Github Action

janpio:


janpio:
/rebase
(will probably not do anything - as the workflow needs to be in main branch)

PR #883, opened 2020-10-14
feat(test): test auto recovery

### LINKED ISSUE #3451: Move full `blog-env-postgresql` to e2e tests
In our integration test suite, we have a test called `blog-env-postgresql`.
That test goes over the timeouts we have built-in and therefore takes a long time to succeed (about 30s).

We should move this full test instead into the e2e tests, which run async from the normal publish process, so we can iterate faster.

It's still valuable to have a minimal postgres test in the integration tests, so this test can be shortened.

eemmiillyy:
Resolves prisma/prisma#3451


.github/workflows/test.yaml
```diff
@@ -68,7 +68,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        feature: [auto-reconnect]
+        feature: [auto-reconnect, blog-env-postgresql]
+    services:
+      postgres_isolated:
```
eemmiillyy: Perhaps this should be renamed to 'postgres', 'pg_core_features_blog_env_docker' or something in between?

core-features/blog-env-postgresql/index.spec.js
```diff
@@ -0,0 +1,160 @@
+const assert = require('assert')
+const getPort = require('get-port')
+const tcpProxy = require('node-tcp-proxy')
+const { Client } = require('pg')
+const { PrismaClient, PrismaClientValidationError } = require('@prisma/client')
+
+describe('Prisma client and postgres works with a proxy and flakyness', () => {
+  let requests = []
```
eemmiillyy: `let` could be replaced with `const` I think?

core-features/blog-env-postgresql/index.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  let errorLogs = []
```
eemmiillyy: Same comment as above(?)

core-features/blog-env-postgresql/index.spec.js
divyenduz: If it doesn't change then yes

core-features/blog-env-postgresql/index.spec.js
divyenduz: If it doesn't change then yes

core-features/blog-env-postgresql/index.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  afterAll(async () => {
+    await prismaClient.$disconnect()
+    await client.query(`
+    DROP TABLE IF EXISTS "public"."Post" CASCADE;
+    DROP TABLE IF EXISTS "public"."User" CASCADE;`)
+    await client.end()
+  })
+
+  it('should verify that requests are not made after prisma client has disconnected', async () => {
+    const proxy = tcpProxy.createProxy(newPort, hostname, port, {})
```
divyenduz: @timsuchanek Do you know if we still need this proxy? vs using the database directly?

divyenduz (approved):
Besides the proxy and local docker being different from other tests set up, this looks good to me. 

If we can use the remote DB and remove the proxy (Emily informed me that there is an issue with using proxy and remote DB together), we should do it for consistency.

.github/workflows/test.yaml
```diff
@@ -68,7 +68,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        feature: [auto-reconnect]
+        feature: [auto-reconnect, blog-env-postgresql]
```
janpio: `blog-env-postgresql` is not a core feature. 

What is this actually testing and why?

janpio (changes requested):

.github/workflows/test.yaml
eemmiillyy: This is testing that the query engine recovers after an error is thrown. Hence the 16 second time out on line 150 of index.spec.js. The prisma client is going through a proxy to get to the database, so the query on line 129 should fail as the port is not yet open. Queries made after the proxy is "created" and > 15 seconds have elapsed since the last bad query was made should pass (afaik). 

The logic is from this old test:
https://github.com/prisma/prisma/blob/6b97cf3498116d7b5cac37a9a169bbd8e08685d3/src/packages/client/src/__tests__/runtime-tests/blog-env-postgresql/index.js

But as your question suggests this is really not apparent from the naming so I should change that and also remove anything that doesn't directly test that.

janpio: That would be neat 👍

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+    await prismaClient.$disconnect()
+    await client.query(`
+    DROP TABLE IF EXISTS "public"."Post" CASCADE;
+    DROP TABLE IF EXISTS "public"."User" CASCADE;`)
+    await client.end()
+  })
+
+  it('should verify that the prisma client automatically recovers after an interrupted db server connection', async () => {
+    const proxy = tcpProxy.createProxy(newPort, hostname, port, {})
+    await prismaClient.$connect()
```
janpio, line 93:
```suggestion

    // confirm connection is working generally
    await prismaClient.$connect() 
```

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      await prismaClient.user.findMany()
+    } catch (e) {
+      if (!(e instanceof PrismaClientValidationError)) { 
+        throw new Error(`Validation error is incorrect`)
+      }
+      errorLogs.push(e)
+    }
+    expect(errorLogs.length).toBe(0)
+    await prismaClient.$disconnect()
```
janpio, line 106:
```suggestion

    // reset engine and connection
    await prismaClient.$disconnect()
```

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+    await prismaClient.$connect()
+    proxy.end()
```
janpio, line 110:
```suggestion

    // confirm query fails when connection is taken away
    proxy.end()
```

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      await prismaClient.user.findMany()
+    } catch (e) {
+      if (!(e instanceof PrismaClientKnownRequestError)) { // should throw P1001 - Can't reach database server
```
janpio: this should probably be tested explicitly?

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      await prismaClient.user.findMany()
+    } catch (e) {
+      if (!(e instanceof PrismaClientKnownRequestError)) { // should throw P1001 - Can't reach database server 
+        throw new Error(`Error code is incorrect`)
+      }
+      errorLogs.push(e)
+    }
+    expect(errorLogs.length).toBe(1)
+    const proxy2 = tcpProxy.createProxy(newPort, hostname, port, {})
```
janpio, line 119:
```suggestion

    // confirm query works again after creating proxy again and waiting >15s
    const proxy2 = tcpProxy.createProxy(newPort, hostname, port, {})
```

core-features/auto-recovery/index.test.js
```diff
@@ -0,0 +1,130 @@
...
+    await new Promise((r) => setTimeout(r, 16000))
```
eemmiillyy, line 120: I noticed that the timeout here, initially added in consideration of the 15s health check that the connection pooler is supposed to do automatically, is not necessary for the query immediately after the port is reopened to succeed. No new connect() or timeout required. 
Since the timeout is no longer needed, and the fact that we're really only testing prisma client recovery rather than the whole health check process, maybe this was actually fine to live inside the integration test folder after all... Thoughts? @janpio

janpio (commented):
Left a few change suggestions to reflect what I think this does. Correct?

core-features/auto-recovery/index.test.js
eemmiillyy: Yeah I'll match a snap shot instead.

core-features/auto-recovery/index.test.js
janpio: That is interesting - because that means this test is testing something else than what it _should_ test. Probably something to check with the person that wrote the test in the first place.

core-features/blog-env-postgresql/index.spec.js
janpio: The proxy seems to be used to have a way to manually kill the connection - by giving that as the connection string, we control in code if the connection lives or dies. Unless we have another way to control the connection, the proxy is needed.

core-features/auto-recovery/index.test.js
janpio: I might have a hunch why this is working without the timeout - right now it possibly just opens another connection to the database. You might want to limit the amount of connections the Query Engine can open to 1 (connection limit = 1). That should force a reuse of the connection that is now broken and hopefully show the behavior we expect.

janpio:
Current error is: https://github.com/prisma/prisma-client-js/issues/907 - first version this was fixed was `2.9.0-dev.71`.

core-features/auto-recovery/index.test.js
eemmiillyy: Just tried this in my last commit 9cbc6ad and it still succeeds without the timeout :/ (when I run the test locally).
I also moved the test postgres service into the auto-recovery folder for now (just seen that the CI test is failing though - missed something in run.sh apparently).

core-features/blog-env-postgresql/index.spec.js
timsuchanek: @janpio correct!

core-features/auto-recovery/index.test.js
timsuchanek: (I wrote the original test)
I think Jan's theory is correct here, we need to limit the number of connections to 1 by adding `&connection_limit=1` to the connection string.

eemmiillyy:
Note: Upgrading prisma version from `2.9.0-dev.49` to `2.10.0-dev.81` in commit 289974b0a4be335c0a20ff0e2af8f5dd4505d155 appears to change the error message to a less informative one. 

Screenshot 2020-10-28 at 09 45 50

A closed connection to the database now appears to throw error of type `PrismaClientUnknownRequestError2` rather than a validation  error.

Screenshot 2020-10-28 at 10 04 19

eemmiillyy:
Before upgrading versions: https://github.com/prisma/e2e-tests/runs/1319476293 
tests passed with or without a timeout - connection from client pool was automatically valid again as soon as connection to db re-established. 

After upgrading versions: [`8f2947f` (#883)](https://github.com/prisma/e2e-tests/pull/883/commits/8f2947f3630b1d4eb83ae538785f2ec4d5ec2c2f) connection from client pool was not automatically valid again, and adding a timeout of > 15s did not result in a new valid connection being given either. 

Explicit prismaClient.disconnect() then prismaClient.connect() is currently required as shown in commit (918f6f1) and test run https://github.com/prisma/e2e-tests/runs/1320093837

PR #899, opened 2020-10-16
feat(database): mssql basic

### LINKED ISSUE #890: MSSQL
Database tests for MSSQL

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/890


databases/mssql/prisma/schema.prisma
```diff
@@ -0,0 +1,16 @@
+// Like, postgresql://user:password@localhost:5432/database/schema
```
janpio: ?

.github/workflows/test.yaml
```diff
@@ -464,6 +465,7 @@ jobs:
       DATABASE_DO_PG_BOUNCER_URL: ${{ secrets.DATABASE_DO_PG_BOUNCER_URL }}
       HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
       DATABASE_HEROKU_PGBOUNCER_URL: ${{ secrets.DATABASE_HEROKU_PGBOUNCER_URL }}
+      DATABASE_URL_DB_MSSQL: ${{ secrets.DATABASE_URL_DB_MSSQL }}
```
janpio, line 468: What database is this using? I can see no service or docker definition.

databases/mssql/prisma/schema.prisma
divyenduz: From example maybe?

.github/workflows/test.yaml
divyenduz: AWS RDS MSSQL (in 1password already).

databases/mssql/prisma/schema.prisma
divyenduz: Removed

.github/workflows/test.yaml
janpio: Do we usually test external databases for the basic database tests?
If not, we should switch this to Docker as well to not depend on an external service here.
If testing that service is relevant, we can do that as well of course if there is value in it.

divyenduz: Almost (if not all) the tests in this repo so far are on an external Postgres RDS, I just did the same for consistency.

janpio: Ah, we don't actually have tests for pure mysql and postgres - that is done in the integration tests.
All good then - but maybe rename it to include `aws-rds` so it is clear what is being tested.

divyenduz: Renamed to `mssql-aws-rds`, and the test is green.

divyenduz: (firebase-functions failed but that is ofc unrelated to this PR)

PR #1022, opened 2020-11-09
fix(update-script): add peer dep for graphql-request

divyenduz:


package.json
```diff
@@ -3,6 +3,7 @@
     "@prisma/client": "2.11.0-dev.25",
     "arg": "4.1.3",
     "fetch-retry": "4.0.1",
+    "graphql": "15.4.0",
```
janpio, line 6: Ugh is this really the right place to do this?

divyenduz: Why not? In theory, this can be a devDep in this case since we are only using `graphql-request` for test case

PR #1027, opened 2020-11-09
feat(community-generators): new workflow, prisma-dbml-generator test

### LINKED ISSUE #850: Test community generators
- Input on generators to test: https://app.redash.io/prismaio/queries/476639
- Make tests optional or fyi, separate workflow so build status is separate
- (Nexus Plugin Prisma test can also be part of this "fyi" test suite for example)

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/850

- [x] New workflow for optional tests (that fails normally)
- [x] Separate slack webhooks for optional tests
- [x] `prisma-dbml-generator` test


community-generators/prisma-dbml-generator/dbml/schema.dbml
```diff
@@ -0,0 +1,9 @@
+//// ------------------------------------------------------
+//// THIS FILE WAS AUTOMATICALLY GENERATED (DO NOT MODIFY)
+//// ------------------------------------------------------
+
+Table User {
+  id String [pk]
+  email String [unique, not null]
+  name String
+}
```
janpio, line 9: Should this be committed when this is effectively what is compared against in the test? The generator not doing anything would still lead to success this way I guess.

divyenduz: Good point, I can remove it at the beginning of the workflow. 

If `generate` fails the exit code should be non-zero and that should fail the test.

PR #1145, opened 2020-12-07
test(core-feature/browser-build): @prisma/client browser build

### LINKED ISSUE #1101: Create E2E tests for @prisma/client browser build
- Depends on https://github.com/prisma/prisma/pull/4329
- Enum and Type fields should be accessible on the frontend
- Trying to use anything else on the frontend should throw an error 
	**Example**
	`const client = new PrismaClient()`
	**Throws**
	`Error: PrismaClient is unable to be run in the browser.
In case this error is unexpected for you, please report it in https://github.com/prisma/prisma-client-js/issues`

williamluke4:
Fixes #1101


divyenduz (changes requested):
Instead of extending the `nextjs` framework test, a minimal standalone test possibly without any frameworks would be better

divyenduz (approved):
This provides value so we merge it now. Simplification happens later https://github.com/prisma/e2e-tests/issues/1170

.github/workflows/test.yaml
```diff
@@ -325,7 +325,7 @@ jobs:
         id: run-test
         uses: nick-invision/retry@v2
         with:
-          timeout_minutes: 10
+          timeout_minutes: 15
```
divyenduz:
```suggestion
          timeout_minutes: 10
```

PR #1152, opened 2020-12-08
feat: test suite coverage

### LINKED ISSUE #254: GH Action step that checks if all folder are present in Matrix
Right now the effectiveness of our tests depends on https://github.com/prisma/prisma2-e2e-tests/blob/master/.github/workflows/test.yaml actually inlcuding all the subfolders of platforms, bundlers etc in the `matrix` entry. If you create the project but forget to add it there, the tests are not actually run. It might be worth writing a script that can be executed as a step that checks if all the subfolders are present with their own entry in the matrix. If not it could fail the test.

(This would also force us to clean up unused folders etc)

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/254

Instead of a step, I made this into a separate workflow that runs whenever the test workflows run.

This is how a failing case looks like https://github.com/prisma/e2e-tests/runs/1518984260

We actually found an unused folder, in this PR, I added a todo to deal with that separately.

### DIFF
--- a/.github/scripts/check-folders.js
+++ b/.github/scripts/check-folders.js
@@ -18,7 +18,7 @@ async function main() {
         'platforms/firebase-functions/functions', // Firebase root doesn't have package.json but is included
         'generic/basic', // generic/basic doesn't use Github action matrix feature which we parse to find out the differences
         'packagers/yarn-workspaces/prisma-project', // Yarn workspaces root doesn't have package.json but is included
-        // 'platforms/serverless-lambda', // TODO: This should be included, we actually found a missing folder
+        'platforms/serverless-lambda', // TODO: This should be included, we actually found a missing folder
       ]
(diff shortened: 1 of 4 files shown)


janpio (approved):

.github/scripts/check-folders.js
```diff
@@ -0,0 +1,89 @@
...
+  const testYaml = yaml.parse(testYamlString)
+
+  const jobKeys = Object.keys(optionalTestYaml['jobs'])
+    .concat(Object.keys(testYaml['jobs']))
+    .filter((key) => {
+      const jobsToIgnore = [
+        'start-time',
+        'report-to-slack-success',
+        'report-to-slack-failure',
+        'os',
```
janpio: maybe add a comment to each why to ignore?

PR #1173, opened 2020-12-10
chore: enable sls lambda

### LINKED ISSUE #1158: Handle platforms/serverless-lambda
via https://github.com/prisma/e2e-tests/issues/254 we found that `platform/serverless-lambda` is not tested/called. 

This issue tracks cleaning it up.

divyenduz:
Fixes https://github.com/prisma/e2e-tests/issues/1158

### DIFF
--- a/.github/workflows/check-for-update.yaml
+++ b/.github/workflows/check-for-update.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: use node 12
-        uses: actions/setup-node@v1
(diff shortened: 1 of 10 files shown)


.github/workflows/check-for-update.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: use node 12
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2-beta
```
janpio, line 14: Issue to fix this pls when available

PR #1175, opened 2020-12-10
chore: fix puppetter scoping condition

divyenduz:
We only use `puppeteer` in `codesandbox` test so it should be installed only for that. This should increase the speed of all other platform tests.


divyenduz:
Dropping https://github.community/t/matrix-cannot-be-used-in-jobs-level-if/17177/2

PR #1318, opened 2021-01-28
fix: use new prisma cli npm package

divyenduz:


community-generators/typegraphql-prisma/generated/typegraphql-prisma/enhance.js
```diff
@@ -0,0 +1,104 @@
+"use strict";
```
janpio, line 1: intentional to commit these files?

divyenduz: Yes, their example does that as well. We can remove it in this context as they can mislead. Separate issue/problem though.

divyenduz: https://github.com/prisma/e2e-tests/issues/1319 will pick it after the package change has settled.

PR #1374, opened 2021-02-05
feat(vercel-api): test postinstall

### LINKED ISSUE #1268: Vercel test: Check, if postinstall was added properly
In order for Prisma to work in Vercel, we need to add a `postinstall` hook in Vercel in build time to the user's application.
If that hook is not added, Prisma can't be generated properly.
Right now it silently fails, if it doesn't work.
We should explicitly check for this in the Vercel e2e test, so we can quickly see what's wrong when debugging it.

williamluke4:
closes https://github.com/prisma/e2e-tests/issues/1268


divyenduz (approved):
✅

PR #1414, opened 2021-02-12
ci: New upgrade-branch-to-version workflow and script

janpio:
This PR adds a `upgrade-branch-to-version.yaml` GH Actions workflow that can be triggered externally with a branch name and a Prisma version. It also adds a `upgrade-branch-to-version.sh` script that is triggered by that workflow to do the actual upgrading.

Right now the `upgrade-branch-to-version.sh` is just a slightly modified version of `check-for-update.sh` in that it does not loop for 5 minutes - but the actual updating logic is unchanged, so the supplied Prisma version from the workflow is **not** actually used but the script still does `npm info` etc to get the version to upgrade to. This should be fixed later.


divyenduz (approved):
LGTM

PR #1417, opened 2021-02-12
fix: undo renovate experiment

janpio:


divyenduz (approved):
✅ 

Let's wait for tests to be green

PR #1438, opened 2021-02-16
ci: Reintegrate Codesandbox test into platforms block

janpio:
Only install Puppeteer when the Matrix is running for Codesandbox, hence we can put them together again and have less jobs.

The missing check is expected, as this will be reconfigured on the repo after this has been merged.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -413,15 +413,15 @@ jobs:
         with:
           node-version: 12
 
+      - name: Install Dependencies
+        run: yarn
+        
       # Install Puppeteer for Codesandbox test only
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -462,6 +415,13 @@ jobs:
 
       - name: Install Dependencies
         run: yarn
+        
+      # Install Puppeteer for Codesandbox test only
+      - name: Install Puppeteer
+        if: ${{ matrix.platform == 'codesandbox' }}
```
timsuchanek, line 421: Nice!!

timsuchanek (approved):

PR #1498, opened 2021-03-05
chore(ci): Simplify Slack part

janpio:
This quickly simplifies the (repeated) Slack jobs in 2 workflows.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,6 +2,8 @@ name: test
 on: push
 env:
   PRISMA_TELEMETRY_INFORMATION: 'e2e-tests test.yaml'
+  SLACK_WEBHOOK_URL_WORKFLOWS: ${{ secrets.SLACK_WEBHOOK_URL_WORKFLOWS }}
+  
 jobs:
   start-time:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -30,19 +32,9 @@ jobs:
     if: success() && (github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/integration' || github.ref == 'refs/heads/patch-dev' || github.ref == 'refs/heads/latest')
     steps:
       - uses: actions/checkout@v2
-
-      - name: Install Dependencies
-        run: yarn install
```
janpio, line 35: This is not needed, the shell script installs its own dependencies

.github/workflows/test.yaml
```diff
@@ -30,19 +32,9 @@ jobs:
...
-
       - uses: actions/setup-node@v2
-        with:
-          node-version: 12
```
janpio, line 39: No dependency to Node 12, should work with stable all the time

.github/workflows/test.yaml
```diff
@@ -30,19 +32,9 @@ jobs:
...
-
       - name: report success to slack
         run: bash .github/scripts/slack-workflow-status.sh ":white_check_mark:"
-        env:
-          CI: 1
-          SLACK_WEBHOOK_URL_WORKFLOWS: ${{ secrets.SLACK_WEBHOOK_URL_WORKFLOWS }}
```
janpio, line 45: Not necessary here as defined globally

Jolg42 (approved):
Nice and clean! 🧹

PR #1510, opened 2021-03-12
fix(platform/lambda): remove @prisma/engines (cli binaries) from zip

matthewmueller:
- ~~only use binaryTarget native~~
- remove @prisma/engines from zip

### DIFF
--- a/platforms/lambda/prisma/schema.prisma
+++ b/platforms/lambda/prisma/schema.prisma
@@ -9,7 +9,7 @@ datasource db {
 }
 
 model User {
-  id    String  @default(cuid()) @id
+  id    String  @id @default(cuid())
   email String  @unique
   name  String?
 }
--- a/platforms/lambda/zip.sh
+++ b/platforms/lambda/zip.sh
@@ -8,6 +8,7 @@ yarn install --production
 rm -rf lambda.zip
 
 rm -rf node_modules/prisma
+rm -rf node_modules/@prisma/engines
 rm -rf node_modules/typescript
 
 zip -r lambda.zip index.js prisma/schema.prisma node_modules/.prisma node_modules/**


platforms/lambda/prisma/schema.prisma
```diff
@@ -1,6 +1,6 @@
 generator client {
   provider      = "prisma-client-js"
-  binaryTargets = ["native", "rhel-openssl-1.0.x"]
+  binaryTargets = ["native"]
```
Jolg42: `"rhel-openssl-1.0.x"` is needed for Lambda

See https://www.prisma.io/docs/guides/deployment/deploying-to-aws-lambda#binary-targets-in-schemaprisma

Jolg42: native is to make it works on your local machine

matthewmueller: Why do we end up with `query-engine-debian-openssl-1.1.x` then? I figured lambda' `native` is `query-engine-debian-openssl-1.1.x`

Jolg42: so we end up with 

From .prisma/client (to use for @prisma/client)
```
  adding: node_modules/.prisma/client/query-engine-rhel-openssl-1.0.x (deflated 64%)
  adding: node_modules/.prisma/client/query-engine-debian-openssl-1.1.x 
```

From @prisma/engines (binaries for prisma cli only)
```
  adding: node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x (deflated 65%)
  adding: node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x (deflated 64%)
  adding: node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x (deflated 70%)
  adding: node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x (deflated 63%)
```

Jolg42: Here the CI native is ubuntu which is Debian

Jolg42:
```suggestion
  binaryTargets = ["native", "rhel-openssl-1.0.x"]
```
This is needed so generate works on native (in CI) and on "rhel-openssl-1.0.x" which is lambda

matthewmueller: Ah, makes sense! Fixing!

Jolg42: Just fixed in a comment 😄

matthewmueller: Oh hah, woops. My github workflows are a bit dated 😅

PR #1513, opened 2021-03-15
fix(heroku): Replace snap install of Heroku CLI with npm

janpio:
This moves the Heroku CLI installation from snapcraft to npm. Heroku does not recommend that, but only because of the auto updater that is missing for npm. As we are installing with each CI run anyway that is not relevant, instead we have an explicit, versioned dependency on the CLI via npm that is updated via renovate etc.

image


Jolg42 (approved):
Looks good!

PR #1514, opened 2021-03-16
ci: Test specific Node versions

janpio:
Add minimal node version (10.4) and minimal minor versions for Node 12 (12.2) and 14 (14.0) to test matrix.
Also adds latest Node major.


Jolg42 (approved):
Nice!

PR #1515, opened 2021-03-16
ci: Use Node 14 for platforms

janpio:
To see if Firebase works with Node 14 then, which would make https://github.com/prisma/e2e-tests/pull/1463 merge-able.


Jolg42:
And it works!

PR #1570, opened 2021-03-31
fix: update yarn2 berry .cjs file

Jolg42:
File from https://raw.githubusercontent.com/yarnpkg/berry/master/packages/berry-cli/bin/berry.js

Url found in https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/policies.js#L123


janpio:
Why are we including this file in the repository here?

Jolg42:
Removing the file in https://github.com/prisma/e2e-tests/pull/2083

PR #1591, opened 2021-04-07
feat: databases/mongodb-atlas

janpio:
Part of https://github.com/prisma/prisma/pull/6268

Feel free to merge dear reviewers.


databases/mongodb-atlas/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "mongodb-atlas",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "devDependencies": {
+    "mongodb": "^3.6.6",
```
Jolg42, line 7: So it seems we require mongo dep only to create an ObjectID like
```
  const ObjectID = require('mongodb').ObjectID;
  const objectId = new ObjectID().toString();
```

I think it's better if we can remove that dependency in the future

janpio: Prisma currently does not offer this functionatliy afaik :/

Jolg42: I guess not yet indeed

PR #1597, opened 2021-04-07
feat: databases/cockroach-cloud

janpio:
Proof of concept PR that Prisma Client works with Cockroach Cloud.

Attention: We know that Introspection and Migration Engine are not compatible with Cockroach.


millsp (approved):
Thanks!

PR #1613, opened 2021-04-10
feat: Test for deployed binaries

janpio:
This modifies the check script to also expect and compare a "binaryString" that is basically a list of the files in a specific folder in the deployment, `node_modules/.prisma/client`. Test can then define what files (and binaries) to expect there, and fails if this changes.


platforms-serverless/netlify-ci/finally.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+echo "TODO: Read Netlify logs (when available via https://github.com/netlify/cli/issues/119 )"
```
Jolg42, line 3: I upvoted the issue 👍

platforms-serverless/netlify-cli/prisma/schema.prisma
```diff
@@ -6,7 +6,7 @@ datasource db {
 
 generator client {
   provider      = "prisma-client-js"
-  binaryTargets = ["native", "rhel-openssl-1.0.x"]
+  binaryTargets = ["rhel-openssl-1.0.x"]
   output        = "../functions/generated/client"
```
Jolg42, line 10: I think we should test closer to what users are doing by default, meaning we could remove 
`output        = "../functions/generated/client"`

utils/fetch-retry-and-confirm-version.ts
```diff
@@ -1,8 +1,8 @@
 import arg from 'arg'
 import originalFetch from 'node-fetch'
 
-function getExpectedData(prismaVersion: string) {
-  return `{"version":"${prismaVersion}","createUser":{"id":"12345","email":"alice@prisma.io","name":"Alice"},"updateUser":{"id":"12345","email":"bob@prisma.io","name":"Bob"},"users":{"id":"12345","email":"bob@prisma.io","name":"Bob"},"deleteManyUsers":{"count":1}}`
+function getExpectedData(prismaVersion: string, binaryString = "") {
+  return `{"version":"${prismaVersion}","createUser":{"id":"12345","email":"alice@prisma.io","name":"Alice"},"updateUser":{"id":"12345","email":"bob@prisma.io","name":"Bob"},"users":{"id":"12345","email":"bob@prisma.io","name":"Bob"},"deleteManyUsers":{"count":1}${binaryString}}`
```
Jolg42, line 5: >This modifies the check script to also expect and compare a "binaryString" that is basically a list of the files in a specific folder in the deployment, node_modules/.prisma/client. Test can then define what files (and binaries) to expect there, and fails if this changes.

Looks like getExpectedData will never fail since it expects the same var? @janpio

millsp: Isn't that the job of `JSON.stringify(data) !== JSON.stringify(expectedData)` ?

millsp (changes requested):
Changing `rhel-openssl-1.0.x` to `native` is probably better but not compulsory since it is the only binary supported by Netlify + it's hard-coded.

utils/fetch-retry-and-confirm-version.ts
janpio: Yep, @millsp got that right - this just constructs the string that is expected. (Just piling on the current existing code here... will need to be refactored sooner or later.

platforms-serverless/netlify-cli/prisma/schema.prisma
janpio: This is in the `netlify-cli` project, where a complete function directory is manually uploaded with the CLI `deploy`. I don't think without specifying an output directory will work there.

janpio:
What does this comment refer to @millsp? 

The `netlify-cli` project probably needs to have the binary target defined, as there is never any `generate` logic executed on Netlify if you use `deploy` from their CLI to upload the function.

platforms-serverless/vercel-cli/api/index.js
```diff
@@ -35,13 +32,28 @@ export default async (req, res) => {
 
   const deleteManyUsers = await client.user.deleteMany()
 
+  /*
```
janpio, line 35:
```suggestion
  /*
  // list all files deployed in Lambda to debug when tests are failing
```

platforms-serverless/vercel-node-builder/index.js
```diff
@@ -39,14 +37,29 @@ app.get('/', async (req, res) => {
 
   const deleteManyUsers = await client.user.deleteMany()
 
+  // list all files in node_modules/.prisma/client
+  const fs = require('fs')
+  const files = fs.readdirSync(process.env.LAMBDA_TASK_ROOT + "/node_modules/.prisma/client")
+
+  /*
```
janpio, line 44:
```suggestion
  /*
  // list all files deployed in Lambda to debug when tests are failing
```

platforms-serverless/netlify-ci/functions/index.js
```diff
@@ -33,6 +33,16 @@ exports.handler = async function (event, context, callback) {
 
   const deleteManyUsers = await client.user.deleteMany()
 
+  /*
```
janpio, line 36:
```suggestion
  /*
  // list all files deployed in Lambda to debug when tests are failing
```

Jolg42 (approved):
I guess you can merge after resolving conflict 👍

millsp (approved):

PR #1616, opened 2021-04-12
chore: split platforms-serverless from platforms

janpio:
Simple PR that splits serverless platforms from the more serverful ones. Makes it easier to understand what is actually being tested.

(Will need to update required checks of the PR/branch after this is merged so PR checks are correct again.)


Jolg42 (approved):
Looks good, unfortunately, it has now merge conflicts, could you fix that and merge @janpio ?

janpio:
Yep, weird that GH could not fix those itself - it was just the `dev` version updates. Local git had no problem at all to rebase.

PR #1637, opened 2021-04-19
feat(serverless-framework-lambda): Test for deployed binaries, upload generated archive and rename project

### LINKED ISSUE #1611: Serverless: Add code to confirm uploaded binaries
https://www.prisma.io/docs/guides/deployment/deploying-to-aws-lambda#binary-targets-in-schemaprisma

We should add code to the function we upload that confirms only the correct QE binary is uploaded and used.

### LINKED ISSUE #1632: Rename `serverless-lambda` to `serverless-framework-lambda`
to clarify what this is

janpio:
related to https://github.com/prisma/prisma/issues/6644
closes https://github.com/prisma/e2e-tests/issues/1632
closes https://github.com/prisma/e2e-tests/issues/1611
related to https://github.com/prisma/prisma/issues/6644

- Test for deployed binaries
- Upload packaged archive as artifact so it can be inspected manually
- Rename complete project to better reflect what it is


millsp (approved):

platforms-serverless/serverless-framework-lambda/serverless.yml
```diff
@@ -15,3 +15,9 @@ functions:
     name: e2e-tests-serverless-lambda
     environment:
       SERVERLESS_LAMBDA_PG_URL: ${env:SERVERLESS_LAMBDA_PG_URL}
+
+# only include the Prisma binary required on AWS Lambda while packaging
+package:
+  patterns:
+    - '!node_modules/.prisma/client/query-engine-*'
+    - 'node_modules/.prisma/client/query-engine-rhel-*'
```
Jolg42: `query-engine-rhel-*` means that if the current binary target is `rhel-openssl-1.1.x` (1.1) for example you are developing on Linux then both 1.0 and 1.1 would be uploaded to Lambda.

I think Lambda only requires `rhel-openssl-1.0.x`?

platforms-serverless/serverless-framework-lambda/test.ts
```diff
@@ -1,15 +1,16 @@
 import { Prisma } from '@prisma/client'
 import { invokeLambdaSync } from './utils'
 
-const name = 'e2e-tests-serverless-lambda'
+const name = 'e2e-tests-serverless-lambda' // TODO rename to match project name again
```
Jolg42:
```suggestion
const name = 'e2e-tests-serverless-lambda' // TODO rename to match project name again in AWS account
```

Jolg42 (approved):
Nice!
See comment about patterns before merging

platforms-serverless/serverless-framework-lambda/serverless.yml
janpio: True. Is that something we want to codify in these rules though? Seems like we have to, otherwise we could end up with 2 different binaries deployed again :/

Jolg42:
```suggestion
    - 'node_modules/.prisma/client/query-engine-rhel-openssl-1.0.x'
```
I think it's better to be safe, this should do it

Jolg42:
You will need to merge and bypass the tests (not wait for green status)

janpio:
I know, I was just experimenting if I can make Github understand that this is not a bad merge conflict but really it just not understanding where this file moved.

millsp (approved):

platforms-serverless/serverless-framework-lambda/run.sh
```diff
@@ -3,9 +3,13 @@
...
 
 yarn install
-
 yarn prisma generate
-
 yarn tsc
 
+# only package
+yarn sls --version
+yarn sls package
```
millsp: Is this to check what would be packaged by `deploy` ?

janpio: Yes exactly, this creates the package locally in .serverless _and leaves it there_ so we can upload it as an artifact later. I should add a few more comments here, so people do not use this as a template for their own CI.

platforms-serverless/serverless-framework-lambda/run.sh
```diff
@@ -3,9 +3,13 @@
 set -eux
 
 yarn install
-
 yarn prisma generate
-
 yarn tsc
 
+# only package
```
janpio:
```suggestion
# only package, so we can inspect the package that will be uploaded by `deploy` in our CI
# this is not needed if you are a normal serverless framework user - just doing the `package and deploy` below is enough
```

millsp (approved):

PR #1645, opened 2021-04-21
test: add Node 16

Jolg42:
v16
Initial Release: 2021-04-20
Active LTS Start: 2021-10-26
Maintenance LTS Start: 2022-10-18
End-of-life: 2024-04-30

See https://nodejs.org/en/about/releases/


janpio:
Let's keep updating the dev version in that project then a bit, until it hopefully passes and then we add as optional similar to 15 👍

Jolg42:
And it's now green with 2.22.0-dev.25 🎊!

PR #1673, opened 2021-04-29
fix(azure): improve function creation logic

janpio:
This cleans up the function creation logic to set the env vars directly after functionapp creation, and only waits before triggering the function upload. Also renamed the function template folder instead of copying so it only deploys 1 function each time, not 2.


Jolg42 (approved):
Looks good

PR #1689, opened 2021-05-07
feat(vercel-with-nextjs-webpack5): Add Webpack 5 config to Nextjs project deployed to Vercel

janpio:


janpio:
Webpack 5 is already default in new Next versions.

PR #1695, opened 2021-05-12
feat(vercel): Use new NEXT_DEBUG_FUNCTION_SIZE env var

janpio:
TODO: This currently does not work, and the env var is set via web UI at Vercel.
Update: Might actually only work for Nextjs projects in the first place.


janpio:
TODO: Redo as new PR

PR #1710, opened 2021-05-25
feat: PlanetScale in serverless environment

janpio:


janpio:
Obsolete now with availability of connection strings.

PR #1717, opened 2021-05-28
chore: enable napi tests (where possible)

janpio:
- Copy relevant bits from https://github.com/prisma/e2e-tests/pull/1534
- then disable all the (currently) failing tests
- fix ENV var setting for Windows
- Adds script to confirm that all tests use the correct engine type in shell script (locally in CI)
- Adds OS as matrix to all tests so it is output with the checks

Follow up work: https://github.com/prisma/e2e-tests/issues/1731 + https://github.com/prisma/e2e-tests/issues/1729


package.json
```diff
@@ -1,6 +1,5 @@
 {
   "dependencies": {
-    "@prisma/client": "2.24.0-dev.27",
```
janpio: This is a change I made to get rid of unexpected versions and Prisma installations being there. 
No idea what the side effects could be or where this might be used.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "@prisma/client": "2.24.0-dev.38",
```
janpio: These were removed intentionally in this branch @williamluke4 as it can influence the available dependencies in the projects we are testing which can be a problem when looking for binaries etc.

.github/scripts/check-client-binaries.sh
```diff
@@ -1,66 +1,70 @@
+echo "Checking Generated Client QE Binary "
+
+dir=$1
+project=$2
+skipped_projects=(prisma-dbml-generator prisma-json-schema-generator napi-preview-feature pkg aws-graviton firebase-functions)
```
janpio: Why are these being skipped here? Needs at least a comment

.github/scripts/check-cli-binaries.sh
```diff
@@ -1,7 +1,14 @@
-echo "Checking CLI/Engines QE Binary "
+echo "-------------- Checking CLI/Engines QE Binary --------------"
 DIR=$1
 PROJECT=$2
 
+skipped_projects=(aws-graviton firebase-functions)
```
janpio, line 7: why are these skipped here? needs at least a comment.

williamluke4: Added Comments

.github/scripts/check-client-binaries.sh
williamluke4: Added Comments

.github/scripts/check-cli-binaries.sh
```diff
@@ -0,0 +1,77 @@
+echo "-------------- Checking CLI/Engines QE Binary --------------"
+DIR=$1
+PROJECT=$2
+
+# These are skipping because they have different project structures
```
janpio, line 5:
```suggestion
# These are skipping because they have different project structures
# TODO Adapt tests so they also work here, or adapt project to fit into the mold
```

.github/scripts/check-client-binaries.sh
```diff
@@ -0,0 +1,75 @@
+echo "-------------- Checking Generated Client QE Binary --------------"
+
+dir=$1
+project=$2
+
+# These are skipping for a variaty of reasons like:
+# - Custom project structure
+# - Custom output location
+# - They do not generate a client
```
janpio, line 9:
```suggestion
# - Custom project structure
# - Custom output location
# - They do not generate a client
# TODO Adapt tests so they also work here, or adapt project to fit into the mold
```

.github/scripts/test-project.sh
```diff
@@ -73,6 +76,16 @@ if [ $code -eq 0 ]; then
...
 
+# confirm existence of correct engine
+if [ $code -eq 0 ]; then
+  echo "-------------- Checking Binaries ---------------"
+  bash ../../.github/scripts/check-cli-binaries.sh $dir $project
+  bash ../../.github/scripts/check-client-binaries.sh $dir $project
+  echo "------------------------------------------------"
+fi
+
+# TODO parse output of npx prisma -v --json for correct file/path
```
janpio, line 87: Didn't we already have that along the way?

williamluke4: Yes but that is part of an individual test (core-features/napi-preview-feature) which uses ts-node and jest and not a bash script

williamluke4 (approved):

.github/scripts/test-project.sh
janpio: Ah true. Thanks for the reminder.

PR #1721, opened 2021-05-31
feat: test nApi preview feature

janpio:
The nApi preview feature is tested via all other tests, but not by itself yet.
This fixes that.


core-features/napi-preview-feature/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+This tests that the n-api library is used in the generated client and that it works
+
+- Builds the required schema
+- Runs Yarn Install
+- Snapshots files in ./node_modules/@prisma/engines and ./node_modules/prisma
+- Runs Generate
+- Snapshots ./node_modules/.prisma/client
+- Runs some basic tests using the generated client
+- Snapshots `prisma -v`
```
janpio:
```suggestion
- Builds the required schema
- Runs `yarn install`
- Snapshots files in `./node_modules/@prisma/engines` (CLI Engines) and `./node_modules/prisma` (CLI)
- Runs `prisma generate`
- Snapshots `./node_modules/.prisma/client` (Generated client with engine)
- Runs some basic tests using the generated client
- Snapshots `prisma -v`
```

core-features/napi-preview-feature/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,324 @@
...
+@prisma/client       : placeholder
+Current platform     : placeholder
+Query Engine         : placeholder
+Migration Engine     : placeholder
+Introspection Engine : placeholder
+Format Binary        : placeholder
+Default Engines Hash : placeholder
+Studio               : placeholder
+Preview Features     : placeholder"
+`;
```
janpio: Ok, this is only testing for "Query Engine" vs "N-API Query Engine", but not the path that is being used.

Can we add that somehow?

core-features/napi-preview-feature/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,324 @@
...
+Studio               : placeholder"
+`;
+
+exports[`N-API Off: ./node_modules/.prisma/client 1`] = `
+Array [
+  "index-browser.js",
+  "index.d.ts",
+  "index.js",
+  "package.json",
+  "query-engine-debian-openssl-1.1.x",
```
janpio, line 52: Simple way to test this on other platforms? Can we somehow have snapshots that depend on the OS this is running on? (We have some OS tests... maybe those help here somehow to figure this out.)

janpio (commented):
We should also add a test where we set a `binaryTargets` in the generator so we see how that turns out. Can be via a separate, follow up PR though I would say so we can get this merged.

core-features/napi-preview-feature/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+async function install(env?: Record<string, string>) {
+  await execa('yarn', ['install', '--force'], {
+    ...defaultExecaOptions,
+    env,
+  })
+}
+async function version(env?: Record<string, string>) {
+  const result = await execa('prisma', ['-v'], {
```
janpio: I think this is run via `yarn` - then adding  `-s` to that command might help to get more stable output (and hence snapshots) (even if Yarn versions change in the future).

core-features/napi-preview-feature/prisma/schema.prisma
```diff
@@ -0,0 +1,23 @@
+
+datasource db {
+  provider = "sqlite"
+  url      = "file:./dev.db"
+}
+generator client {
+  provider        = "prisma-client-js"
+  binaryTargets = ["native"]
```
janpio: setting "native" might have different effects than not setting anything maybe? so probably worth _another_ test or leaving it off here to be more what a normal user would do.

core-features/napi-preview-feature/__snapshots__/test.ts.snap
williamluke4: Added

core-features/napi-preview-feature/utils.ts
williamluke4: Added

core-features/napi-preview-feature/prisma/schema.prisma
williamluke4: Removed

core-features/napi-preview-feature/__snapshots__/test.ts.snap
williamluke4: I think the easiest way to do this would be to generate an object containing the expected files and switch the binary depending on the platform. We would also have to account for binary targets as these would add additional binaries.

Jolg42:
Do we want to merge this before or after 2.24 release?

core-features/napi-preview-feature/__snapshots__/test.ts.snap
janpio: That sounds like a plan. Can you reflect that in an issue maybe? Or at least in the Napi Notion document list?

PR #1734, opened 2021-06-03
fix: Remove top level dependencies

janpio:


millsp:
Once https://github.com/prisma/prisma/pull/7834 is merged, well need to re-adapt `core-features/client-install-on-generate` to have a parent folder with `node_modules` with `prisma` & `@prisma/client` installed and a sub folder with `node_modules` and no `@prisma/client` (the actual test). This way we can keep things the same testing logic after merging *this* PR.

millsp (approved):

janpio:
> well need to re-adapt `core-features/client-install-on-generate` to have a parent folder with `node_modules` with `prisma` & `@prisma/client` installed and a sub folder with `node_modules` and no `@prisma/client` (the actual test).

Why though? This test fundamentally should just make sure that the Client is installed during `generate` when it is missing. Wouldn't that adapted test test a special case of that?

janpio:
That totally did not work, the next commit already created the dependency again :( 
https://github.com/prisma/e2e-tests/commit/4bc1ee99e549c1f60485328f8bc20be0e73f0cb8
Must have missed something.

(Or one of the scripts was still running... will try to kill them and delete deps again)

janpio:
Seems it was the already running script - now it seems fine and persist.

PR #1738, opened 2021-06-03
chore: skip lint when auto sync bump

Jolg42:
[skip ci]


Jolg42:
@janpio this could save some runs too.

I think we only need this lint when a human does a PR/push basically?

janpio (approved):
Yep, looks good. Renovate should not mess this up :D

PR #1752, opened 2021-06-06
fix: remove exclusion for Node API `azure-functions-linux`

janpio:
part of https://github.com/prisma/e2e-tests/issues/1748


platforms-serverless/azure-functions-linux/func-placeholder/index.ts
```diff
@@ -1,7 +1,13 @@
...
+import { Prisma, PrismaClient } from '@prisma/client'
 
-const client = new PrismaClient()
+const client = new PrismaClient({
+  datasources: {
+    db: {
+      url: process.env.AZURE_FUNCTIONS_LINUX_PG_URL,
+    },
+  },
+})
```
janpio: We should technically only do this for Node-API.

platforms-serverless/azure-functions-linux/finally.sh
```diff
@@ -1,12 +1,30 @@
...
+
 set -eux
 
 app="$(cat func-tmp.txt)"
 
 group="prisma-e2e-linux"
+subscriptionId="63cbd994-4c44-4b0d-ba9f-463c6ec6ee31"
+
+echo "Too View Logs Visit"
+echo "https://portal.azure.com/#blade/WebsitesExtension/FunctionMenuBlade/monitor/resourceId/%2Fsubscriptions%2F${subscriptionId}%2FresourceGroups%2F${group}%2Fproviders%2FMicrosoft.Web%2Fsites%2F${app}%2Ffunctions%2F${app}"
```
janpio: Will only be available if we do not delete the function and insights, so should be moved down into the  `else`.

williamluke4: Moved to else

platforms-serverless/azure-functions-linux/func-placeholder/index.ts
williamluke4: Removed

williamluke4 (approved):

PR #1754, opened 2021-06-06
feat: add comments where engine file checking should be for remote tests (platforms, platforms-serverless)

janpio:
This PR adds a few comments with TODOs or problems, and cleans up the existing files a tiny bit.

part of https://github.com/prisma/e2e-tests/issues/1729


platforms-serverless/vercel-cli/test.sh
```diff
@@ -2,4 +2,6 @@
 
 set -eu
 
+# Note: THIS ALREADY USES THE binary string, so we can copy this for other tests of the same pattern!
+# TODO This is not dynamic based on Napi env var, so should technically not even work... wtf?
```
janpio: This should fail on Node-API.
Probably does not because of binaryTargets?

platforms-serverless/vercel-node-builder/test.sh
```diff
@@ -2,4 +2,6 @@
 
 set -eu
 
+# Note: THIS ALREADY USES THE binary string, so we can copy this for other tests of the same pattern!
+# TODO This is not dynamic based on Napi env var, so should technically not even work... wtf?
```
janpio: This should fail on Node-API.
Probably does not because of binaryTargets?

williamluke4 (approved):

PR #1776, opened 2021-06-09
chore: drop Node.js 10 support

millsp:
Related https://github.com/prisma/prisma/pull/7573


.github/workflows/test.yaml
```diff
@@ -335,7 +333,7 @@ jobs:
       - name: use node ${{ matrix.node }}
         uses: actions/setup-node@v2
         with:
-          node-version: 10
+          node-version: ${{ matrix.node }}
```
janpio: there is no matrix with node versions in this job, the name is just a bug here I would say

.github/workflows/test.yaml
```diff
@@ -380,7 +378,7 @@ jobs:
...
```
janpio: there is no matrix with node versions in this job, the name is just a bug here I would say

libraries/apollo-server/package.json
```diff
@@ -19,6 +19,6 @@
     "typescript": "4.2.4"
   },
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=12.2.0"
```
janpio: why do we even need this in these generic `libaries` projects?

millsp: Probably a mistake, I'll remove them

millsp:
@janpio any idea about netlify failure https://github.com/prisma/e2e-tests/runs/2788317219#step:5:413?
Actually, I think that this could be related to my [PR](https://github.com/prisma/prisma/pull/7111), I need to investigate

janpio:
PR failure completely unrelated and fixed via https://github.com/prisma/e2e-tests/pull/1778 - sorry you got distracted by that stupid thing.

millsp:
~~`test / platforms-napi (heroku, ubuntu-latest) (push)` is failing. @williamluke4 do you know why?~~ It's flaky

.github/scripts/test-project.sh
```diff
@@ -37,7 +37,8 @@ if [ -f "prepare.sh" ]; then
   echo "prepare script found, executing $dir/$project/prepare.sh"
   echo ""
 
-  bash prepare.sh
+  # execute & allow export of env vars
+  . prepare.sh
```
janpio, line 41: Uh, what does this do and how?

.github/workflows/test.yaml
```diff
@@ -217,14 +217,12 @@ jobs:
       matrix:
         engine: ['napi', 'binary']
         node:
-          - 10
           - 12
           - 14
-          - 10.16.0 # minimal node version via https://github.com/prisma/prisma/blob/master/src/packages/client/package.json and https://www.prisma.io/docs/reference/system-requirements
```
janpio, line 223: We will need to update our package.json, right? So we should keep the comment on the node version that will apply to (12.2.0 I guess?)

packagers/npm-global/prepare.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+mkdir -p ~/.npm-global
+npm config set prefix '~/.npm-global'
+NPM_CONFIG_PREFIX=~/.npm-global
+PATH=~/.npm-global/bin:$PATH
```
janpio, line 6: Why did this become necessary and what does it really do?

millsp: On Node.js > 10, it does not have the correct permissions to access the global modules folder (needs to be sudo/root, on linux only). It is a known limitation and documented here https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally

.github/scripts/test-project.sh
millsp: Running with `bash` creates a sub-shell that inherits from the parent env. But doing `.` runs it in the same shell, allowing us to add env vars into the parent shell (because shell has env isolation per process). Useful for `prepare` to set some env vars for what happens next.

.github/workflows/test.yaml
millsp: If I understood well, I have done that here https://github.com/prisma/prisma/pull/7573. In here we now have:
```
        node:
          - 12
          - 14
          - 15
          - 16
          - 12.2.0 # minimal minor version of node 12 via https://www.prisma.io/docs/reference/system-requirements
          - 14.0.0 # minimal minor version of node 14 via https://www.prisma.io/docs/reference/system-requirements
```

.github/workflows/test.yaml
```diff
@@ -217,14 +217,12 @@ jobs:
...
-          - 12.2.0 # minimal minor version of node 12 via https://www.prisma.io/docs/reference/system-requirements
-          - 14.0.0 # minimal minor version of node 14 via https://www.prisma.io/docs/reference/system-requirements
           - 15
           - 16
+          - 12.2.0 # minimal minor version of node 12 via https://www.prisma.io/docs/reference/system-requirements
```
```suggestion
          - 12.2.0 # minimal node version via https://github.com/prisma/prisma/blob/master/src/packages/client/package.json and minimal minor version of node 12 via https://www.prisma.io/docs/reference/system-requirements
```

.github/workflows/test.yaml
janpio: Yep, this is what I mean: https://github.com/prisma/e2e-tests/pull/1776/files#r649408707

.github/scripts/test-project.sh
janpio: Do we use this for something already?

millsp: Yep, for the `packagers/npm-global`

janpio (approved):
Looks great. 

Do not forget to set the target branch to `dev` again before merging.

PR #1790, opened 2021-06-11
fix(heroku): no local install and generate, just cloud

janpio:


Jolg42 (approved):
Looks old but there is a conflict to resolve because it's a very old PR

PR #1793, opened 2021-06-11
fix(aws-graviton): Remove copy of package.json

janpio:
Question is why this was there in the first place...


Jolg42:
There is no change anymore, see "files changes" in the PR, after syncing with the `dev` branch, so we can close this

PR #1798, opened 2021-06-11
feat(platform/m1-macstadium): Run tests on M1 Mac

janpio:
via SSH to MacStadium M1

(failing tests were also on `dev`, so can be ignored.)


janpio (commented):
Make sure to review the actual test output please.

platforms/m1-macstadium/code/index.test.js
```diff
@@ -0,0 +1,24 @@
+const { PrismaClient, Prisma } = require('@prisma/client')
+
+const prisma = new PrismaClient()
+
+describe('Prisma', () => {
+  afterAll(() => {
+    prisma.$disconnect()
```
Jolg42, line 7: Could use an await probably

Jolg42 (commented):
Looks ok

platforms/m1-macstadium/code/index.test.js
janpio: True, will fix via unrelated PR.

PR #1803, opened 2021-06-14
test(studio): Mutation & Node-API usage tests for Studio

sdnts:


sdnts:
~I'm realizing there's no actual way to tell if Studio used the library or not. The only thing we can check is if the library was downloaded.~

~This iis because IIUC SDK needs the binary at least once to parse the schema. If the schema says it is supposed to use the library, it will (or it should at least) use it going forward. Same holds for Studio. So all this test does is make sure that libraries are downloaded when they should.~

~I learned this because in a previous commit, I was deleting binaries before the test if `PRISMA_FORCE_NAPI` was set, and then expecting that only libraries exist after the test. But that's not possible, because the SDK needs hthe bianry to parse the schema in the first place.~

I was incorrect.

janpio:
No, the CLI/SDK should already use the engine library when you install it with the env var in the first place. No query engine binary necessary at all.

sdnts:
~@janpio Yeah but if I delete all binaries before the Node-API test starts, how would the SDK know to download the Node-API at all? It needs the binary at least once right?~

~(Look at my `test.sh` file for context)~

Not relevant anymore.

core-features/studio/prisma/schema.prisma
```diff
@@ -11,8 +11,4 @@ model User {
     id    Int    @id
     name  String
     email String
-
-    parentId Int
-    parent   User   @relation(fields: [parentId], references: [id])
-    children User[] @relation("UserToUser")
```
sdnts, line 17: Was just causing problems while deleting etc. I have a test for self-referential models in Studio, so all good

core-features/studio/test.sh
```diff
@@ -1,13 +1,49 @@
 #!/bin/sh
 
-set -eu
+set -e
```
sdnts, line 3: Removed `-u` because we do not want to stop the script for undefined variables (in our case, `PRISMA_FORCE_NAPI` might be undefined)

janpio: We should use a test here that considers that - most of the other scripts that contain `PRISMA_FORCE_NAPI` do that.

core-features/studio/test.ts
```diff
@@ -1,13 +1,18 @@
...
 import { createHash } from 'crypto'
 import fetch from 'node-fetch'
+import { PrismaClient } from '@prisma/client'
 
 const STUDIO_PORT = 5555
 const SCHEMA_HASH = createHash('md5')
   .update(fs.readFileSync('./prisma/schema.prisma'))
   .digest('hex')
 
+const prisma = new PrismaClient()
```
janpio: Having that here has the danger, that the binaries are downloaded for the normal PrismaClient usage instead of for Studio, right?

sdnts: Yes, but the assertions in test.sh ensures nothing extra is downloaded.

Regardless, I got rid of it because it felt wrong. I'm now ousing Studio to `findMany` after a `create`.

core-features/studio/test.sh
```diff
@@ -11,3 +25,26 @@ sleep 3 # Studio takes some time to start up
...
+	# If you are supposed to use libraries, check if binaries were downloaded during tests
+	if [[ ! -z "$(ls -1 node_modules/.prisma/client/query-engine* 2>/dev/null)" ]] ; then
+		echo '[1] QE binary exists when it should not'
+		exit 1
+	fi
+	if [[ ! -z "$(ls -1 node_modules/prisma/query-engine* 2>/dev/null)" ]] ; then
+		echo '[2] QE binary exists when it should not'
+		exit 1
+	fi
+fi
```
sdnts, line 50: Here are some assertions that should check if anything extra was downloaded.

If you look on line 11, I delete binaries if `PRISMA_FORCE_NAPI` is set, and then check again here if they exist. If they do, it means someone downloaded them, and the test fails.

The same happens if `PRISMA_FORCE_NAPI` is not set, libraries are deleted before `jest`, and checked after.

sdnts: Let me know if all paths I check are sufficient, can obviously add other places we expect to find the binary / library.

sdnts:
@janpio This is ready for review! Merge whenever!

core-features/studio/test.sh
janpio: On Windows the paths should be different for the Node-API library, but I assume that is not being tested here so does not affect things.

PR #1807, opened 2021-06-15
feat: test `prisma db seed`

### LINKED ISSUE #1806: Test seeding
Run steps from https://github.com/prisma/prisma/issues/7176 in CI here and make sure data is available afterwards.

janpio:
close #1806


Jolg42:
@janpio Linting fails with 
```
There are some folders in e2e-tests that are not used in test.yml or optional-test.yml workflows
[ 'core-features/db-seed' ]
```

Probably because you create a "migrate" job instead of adding it as a "feature" in the "core-features" job.

Should I move "db-seed" folder to a "migrate" directory to fix this? If I correctly understand how it works.

janpio:
> Should I move "db-seed" folder to a "migrate" directory to fix this? If I correctly understand how it works.

Yes, that should do the job. (I have no idea how `/.github/scripts/check-folders.js` works in detail)

.github/workflows/test.yaml
```diff
@@ -174,6 +182,54 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        engine: ['napi', 'binary']
+        feature: [
+          db-seed
+        ]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    concurrency: ${{ github.run_id }}-migrate
```
janpio: This will cancel all runs except 2 unfortunately. We should probably modify the database URL in the test, so it can create a new DB for each test matrix combination, then no `concurrency` necessary.

Jolg42:
Replaced by https://github.com/prisma/e2e-tests/pull/2048

PR #1824, opened 2021-06-19
chore: move napi-preview-feature into its own folder and expand

janpio:
part of https://github.com/prisma/e2e-tests/issues/1785

Moves the existing core-features/napi-preview-feature test, only tests it once instead of binary/napi, but then runs it on all 3 OSes and snapshots the results - and then fully expands it to cover the binary/library problem extensively.

---

Things to review:
- Start with `test.yaml` changes
- Accept that `custom-engines` is an ugly hack to get custom, working engines easily (as it just copies the content of a current `@prisma/engines` to that folder) - Or is there a better way?
- Then look at the `engines/napi-preview-feature/*.test.ts` and the `utils.ts` it uses. Do all the test descriptions make sense for what is actually being tested?
- Finally review the Snapshot files if what they do fits what the test describes
- Profit


.github/workflows/test.yaml
```diff
@@ -166,6 +166,45 @@ jobs:
...
+        id: run-test
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 10
+          max_attempts: 3
+          command: bash .github/scripts/test-project.sh engines ${{ matrix.feature }}
+
+      - name: notify-slack
+        if: failure()
+        run: bash .github/slack/notify-failure.sh core-features ${{ matrix.feature }}
```
millsp:
```suggestion
        run: bash .github/slack/notify-failure.sh engines ${{ matrix.feature }}
```

millsp:
Maybe we need to add

```ts
      - platforms-napi
      - platforms-serverless-napi
```

to `report-to-slack-failure`

Jolg42:
>Accept that custom-engines is an ugly hack to get custom, working engines easily (as it just copies the content of a current @prisma/engines to that folder) - Or is there a better way?

Adding the binaries to Git seems like a bad idea to me, they are quite big in size and it will make this git repository hard to work with, especially if we update these committed engines for new versions.

I think it would be better to add the binaries to a .gitignore

engines/napi-preview-feature/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export async function runTest(options: {
+  previewFeatures?: string[]
+  binaryTargets?: string[]
+  env?: Record<string, string>
+  env_on_deploy?: Record<string, string>
+}) {
+  // This ensures that if PRISMA_FORCE_NAPI is set for the workflow it is removed before running these tests
+  // TODO Should not be necessary any more
+  if (process.env.PRISMA_FORCE_NAPI === 'true') {
```
Jolg42:
```suggestion
  if (process.env.PRISMA_FORCE_NAPI) {
```
So we always remove it if set?

janpio:
> > Accept that custom-engines is an ugly hack to get custom, working engines easily (as it just copies the content of a current @prisma/engines to that folder) - Or is there a better way?
> 
> Adding the binaries to Git seems like a bad idea to me, they are quite big in size and it will make this git repository hard to work with, especially if we update these committed engines for new versions.
> 
> I think it would be better to add the binaries to a .gitignore

How would the code work then? It needs custom engines (both binary and library) for all platforms for the tests to work.

(We could download them before the tests run, but then these tests itself would rely on the downloading logic that is being tested here)

Jolg42:
The alternative to the custom engines hack would be to call "download" programmatically like done here 
https://github.com/prisma/engines-wrapper/blob/master/packages/engines/src/index.ts#L10-L33

engines/napi-preview-feature/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export async function runTest(options: {
+  previewFeatures?: string[]
+  binaryTargets?: string[]
+  env?: Record<string, string>
+  env_on_deploy?: Record<string, string>
+}) {
+  // This ensures that if PRISMA_FORCE_NAPI is set for the workflow it is removed before running these tests
+  // TODO Should not be necessary any more
```
janpio: Do you agree with this @williamluke4? now that we are only running this test once, not once for `napi` and once for `binary` and do not set the env var in the job any more depending on that.

janpio:
> The alternative to the custom engines hack would be to call "download" programmatically like done here
> [https://github.com/prisma/engines-wrapper/blob/master/packages/engines/src/index.ts#L10-L33](https://github.com/prisma/engines-wrapper/blob/master/packages/engines/src/index.ts?rgh-link-date=2021-06-22T09%3A57%3A59Z#L10-L33)

This is an end to end test. Unless that is available in the public API somehow, that is not possible. Even if it were, see previous comment about using the function under test to get the test data in the first place.

engines/napi-preview-feature/utils.ts
millsp: Yeah I think this is not needed anymore

millsp (commented):
Very nice! Great work. ~~Just wondering why we would need to add all those generated files to git?~~

janpio:
TODO after call with @millsp:
- [x] Replace comitted `custom-engines` with `beforeAll` that creates them once
- [ ] ~~Replace snapshots with inline snapshots that are stored in variables~~ => https://github.com/prisma/e2e-tests/pull/1838

williamluke4 (approved):
Profit :rocket:

PR #1837, opened 2021-06-22
chore: remove prisma from root `node_modules`

millsp:
Fixes https://prisma-company.slack.com/archives/D01SYV72V9T/p1624383670000100

If we delete the `@prisma/client` in our test then it fetches the one at the root. However, that is not needed. The dependency was added to the root `package.json` on `prisma generate` when not found.

### DIFF
--- a/core-features/client-install-on-generate/run.sh
+++ b/core-features/client-install-on-generate/run.sh
@@ -5,4 +5,4 @@ set -eux
 yarn install
 yarn remove @prisma/client
 
-yarn prisma generate
+env DEBUG="prisma:generator" yarn prisma generate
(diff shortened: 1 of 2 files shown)


janpio:
Unfortunately the scripts running here will hate this and add it back :)

I actually created PRs to fix this before: https://github.com/prisma/e2e-tests/pull/1826 (See for https://github.com/prisma/e2e-tests/pull/1826/files#diff-eea9ed37aa166f9b318636a69eda0273fde173a1ceab50c0320aca63f814d3c3 for attempt to fix script) + https://github.com/prisma/e2e-tests/pull/1734 Not sure if that would already be enough, never went deep enough and tried it out :/

PR #1864, opened 2021-06-30
test(pnpm): installation in various setups

millsp:
Add tests for `pnpm` setups to prevent regressions in the future.

Related https://github.com/prisma/prisma/pull/7942

Checks:
- [x] New tests have failed on `prisma/master`
- [x] New tests [have succeeded](https://github.com/prisma/e2e-tests/actions/runs/985325657) on the PR


packagers/pnpm/prisma/schema.prisma
```diff
@@ -0,0 +1,14 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("PACKAGERS_NPM_PG_URL")
```
millsp: I see that each package manager has it's own db. Should we set that up for `pnpm`?

janpio: Yes, that would be necessary - technically and optimally even one for each test. That is a decision we made in the beginning that is not scaling great, but we did not have the time to fix it yet.

Easiest way to get the databases right now is to use the internal database provision tool that @madebysid posted in #general this week, then set the connection strings in the repo secrets (if you can't do that, just DM them to me and I can do that).

packagers/pnpm-workspaces-custom-output/prepare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+mkdir -p ~/.npm-global
+npm config set prefix '~/.npm-global'
+NPM_CONFIG_PREFIX=~/.npm-global
+PATH=~/.npm-global/bin:$PATH
+npm i -g pnpm@latest
```
janpio:
```suggestion
# workaround so `npm i -g` can work
mkdir -p ~/.npm-global
npm config set prefix '~/.npm-global'
NPM_CONFIG_PREFIX=~/.npm-global
PATH=~/.npm-global/bin:$PATH

npm i -g pnpm@latest
```

packagers/pnpm-workspaces-custom-output/workspace/sub-project-1/prisma/schema.prisma
```diff
@@ -0,0 +1,15 @@
+generator client {
+  provider = "prisma-client-js"
+  output = "../node_modules/.prisma/client"
```
janpio, line 3: Wouldn't it be better to use a non standard folder for the custom output?

packagers/pnpm-workspaces-custom-output/workspace/sub-project-1/script.ts
```diff
@@ -0,0 +1,25 @@
+import { PrismaClient } from '.prisma/client'
```
janpio, line 1: This is equivalent to importing from @prisma/client, isn't it?

packagers/pnpm-workspaces-default-output/prepare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+mkdir -p ~/.npm-global
+npm config set prefix '~/.npm-global'
+NPM_CONFIG_PREFIX=~/.npm-global
+PATH=~/.npm-global/bin:$PATH
+npm i -g pnpm@latest
```
janpio:
```suggestion
# workaround so `npm i -g` can work
mkdir -p ~/.npm-global
npm config set prefix '~/.npm-global'
NPM_CONFIG_PREFIX=~/.npm-global
PATH=~/.npm-global/bin:$PATH

npm i -g pnpm@latest
```

packagers/pnpm-workspaces-default-output/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+FILE1=sub-project-1/node_modules/.prisma/client/index.js
```
janpio, line 4: Why only check for one of them here?

packagers/pnpm/prepare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+mkdir -p ~/.npm-global
+npm config set prefix '~/.npm-global'
+NPM_CONFIG_PREFIX=~/.npm-global
+PATH=~/.npm-global/bin:$PATH
+npm i -g pnpm@latest
```
janpio:
```suggestion
# workaround so `npm i -g` can work
mkdir -p ~/.npm-global
npm config set prefix '~/.npm-global'
NPM_CONFIG_PREFIX=~/.npm-global
PATH=~/.npm-global/bin:$PATH

npm i -g pnpm@latest
```

janpio (approved):
Looks good, but see comments.

packagers/pnpm-workspaces-custom-output/workspace/sub-project-1/script.ts
millsp: Not in a symlinked/hoisted/workspace. Because `@prisma/client` sits in the hoisted `node_modules` folder, it will find the default generated client which throws an error message. Those generated clients are found in the sub-projects `node_modules`. However, the other test `pnpm-workspaces-default-output` will let two clients overwrite each other in the hoisted `node_modules` and you can then use `@prisma/client`.

packagers/pnpm-workspaces-default-output/test.sh
millsp: Actually, thanks for pointing this out because it would fail. It should be
```
FILE1=node_modules/.prisma/client/index.js
```
And check that it *does not* exist in the sub-folders. This is something I tried to change before and it caused regressions. Therefore I added a test for it. When no specific location is specified, the client is generated near `@prisma/client`, in this case in the hoisted `node_modules`.

packagers/pnpm-workspaces-custom-output/workspace/sub-project-1/prisma/schema.prisma
millsp: I think that we could add another test. With this one I want to make sure that we would not generate in the root `node_modules`, basically testing that it properly gets generated relatively to where we called the `cli`.

PR #1875, opened 2021-07-01
chore(DATABASE_URL): Completely rework how standard database connection strings work (generated instead of provided)

janpio:
This PR removes the individual database URLs per test, and uses `DATABASE_URL` instead that is generated dynamically in the CI configuration.

image

Current state:
- [x] Several tests do seem to rely on a server/platform side configured ENV var that is now not accepted by the project any more, and the new one is not present: heroku, vercel-*, netlify-*
- [x] `databases/planetscale` uses the default ENV var name which triggers the database creation logic, but of course should use the local env var instead. Solution: https://github.com/planetscale/cli/issues/322 Alternative: Exclude folder/project from logic.

After this is merged, quite a few secrets in this repo can be deleted.

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -32,6 +32,7 @@ echo "running $dir/$project"
 # Find schema, if it contains `env("DATABASE_URL")`, db push that schema to database
 if [[ "$project" = "planetscale" ]]
 then
+  true
   # have to skip planetscale until https://github.com/planetscale/cli/issues/322 is implemented
 else
   schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -30,15 +30,18 @@ echo "running $dir/$project"
 
 
 # Find schema, if it contains `env("DATABASE_URL")`, db push that schema to database
-schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
-if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
-  echo ""
-  echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
-  echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
-  npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
-  echo ""
+if [ "$project" = "planetscale" ]; then
+  # have to skip planetscale until https://github.com/planetscale/cli/issues/322 is implemented
+else
+  schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
+  if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
+    echo ""
+    echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
+    echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
+    npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
+    echo ""
+  fi 
 fi
-
 
 echo "cd $dir/$project"
 cd "$dir/$project"
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -30,7 +30,8 @@ echo "running $dir/$project"
 
 
 # Find schema, if it contains `env("DATABASE_URL")`, db push that schema to database
-if [ "$project" = "planetscale" ]; then
+if [[ "$project" = "planetscale" ]]
+then
   # have to skip planetscale until https://github.com/planetscale/cli/issues/322 is implemented
 else
   schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -28,6 +28,18 @@ echo ""
 echo "-----------------------------"
 echo "running $dir/$project"
 
+
+# Find schema, if it contains `env("DATABASE_URL")`, db push that schema to database
+schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
+if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
+  echo ""
+  echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
+  echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
+  npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
+  echo ""
+fi
+
+
 echo "cd $dir/$project"
 cd "$dir/$project"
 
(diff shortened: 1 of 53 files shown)


.github/workflows/test.yaml
```diff
@@ -8,7 +8,11 @@ on:
 
 env:
   PRISMA_TELEMETRY_INFORMATION: 'e2e-tests test.yaml'
+  CI: 1 
   SLACK_WEBHOOK_URL_WORKFLOWS: ${{ secrets.SLACK_WEBHOOK_URL_WORKFLOWS }}
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  SLACK_WEBHOOK_URL_FAILING: ${{ secrets.SLACK_WEBHOOK_URL_FAILING }}
+  DATABASE_URL_POSTGRES_TEMPLATE: ${{ secrets.DATABASE_URL_POSTGRES_TEMPLATE }}
```
janpio, line 15: putting this on the workflow level makes it unnecessary to repeat it for each job

.github/workflows/test.yaml
```diff
@@ -113,17 +113,25 @@ jobs:
         with:
           node-version: 12
 
+      - name: Set test specific DATABASE_URL
+        run: |
+          string=${{ env.DATABASE_URL_POSTGRES_TEMPLATE }}
+          search=database
+          replace=${{ github.job }}_${{ matrix.feature }}_${{ matrix.os }}_${{ matrix.engine }}
+          replaced=${string/$search/$replace}
+          echo "DATABASE_URL=$replaced" >> $GITHUB_ENV
```
janpio, line 122: This is copied to each job, slightly adapted to use the correct name of the feature key, in this case `feature`. Sometimes it is called differently though.

.github/workflows/test.yaml
```diff
@@ -113,17 +113,25 @@ jobs:
...
         with:
           timeout_minutes: 10
           max_attempts: 3
-          command: bash .github/scripts/test-project.sh process-managers ${{ matrix.feature }}
+          command: bash .github/scripts/test-project.sh ${{ github.job}} ${{ matrix.feature }}
 
       - name: notify-slack
         if: failure()
-        run: bash .github/slack/notify-failure.sh process-managers ${{ matrix.feature }}
+        run: bash .github/slack/notify-failure.sh ${{ github.job}} ${{ matrix.feature }}
```
janpio, line 134: Much more copy/paste friendly when you create an additional job.

frameworks/nestjs/prisma/schema.prisma
```diff
@@ -4,7 +4,7 @@ generator client {
 
 datasource db {
     provider = "postgresql"
-    url      = env("FRAMEWORK_NESTJS_PG_URL")
+    url      = env("DATABASE_URL")
```
janpio, line 7: avoids a clash with the new default logic that runs when the env var is `DATABASE_URL`

libraries/apollo-server/test.sh
```diff
@@ -9,7 +9,7 @@ sleep 10
 
 actual="$(curl 'http://localhost:4000/' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://localhost:4000' --data-binary '{"query":"query {\n  test {\n    result\n  }\n}"}' --compressed)"
 
-expected='{"data":{"test":{"result":"[{\"id\":7,\"email\":\"john@example.com\",\"name\":\"John Doe\"}]"}}}'
+expected='{"data":{"test":{"result":"[{\"id\":1,\"email\":\"john@example.com\",\"name\":\"John Doe\"}]"}}}'
```
janpio, line 12: seems the database previously was in a strange state that cause the ID to always be 7? no idea really.

.github/scripts/test-project.sh
```diff
@@ -28,6 +28,23 @@ echo ""
...
+  # have to skip planetscale until https://github.com/planetscale/cli/issues/322 is implemented
+else
+  schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
+  if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
+    echo ""
+    echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
+    echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
+    npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
+    echo ""
+  fi
```
janpio, line 45: This is the logic that ensures if `DATABASE_URL` is used in the schema, we `db push` that database so it exists.

platforms-serverless/azure-functions-linux/run.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eux
 
+echo "DATABASE_URL:"
+echo "$DATABASE_URL"
```
millsp: Do we still need this?

.github/workflows/test.yaml
```diff
@@ -692,16 +723,11 @@ jobs:
       VERCEL_API_PROJECT_ID: ${{ secrets.VERCEL_API_PROJECT_ID }}
       VERCEL_API_ORG_ID: ${{ secrets.VERCEL_API_ORG_ID }}
```
millsp, line 724: Do you think that we can de duplicate this as well?

millsp (approved):
Awesome work!

platforms-serverless/azure-functions-linux/run.sh
janpio: Oh no, that is leftover from debugging 👍 Will remove.

.github/workflows/test.yaml
janpio: In theory yes, in practice the full platforms-serverless is up for a bigger rework anyway so I would tackle it via that.

.github/scripts/test-project.sh
```diff
@@ -28,6 +28,23 @@ echo ""
...
+then
+  true
+  # have to skip planetscale until https://github.com/planetscale/cli/issues/322 is implemented
+else
+  schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
+  if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
+    echo ""
+    echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
+    echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
+    npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
```
Jolg42, line 43: So here npx prisma should run latest I think? Since the root package has no prisma dependency.

platforms-serverless/azure-functions-linux/run.sh
```diff
@@ -2,6 +2,9 @@
...
+
```
Jolg42:
```suggestion
```

Removal as suggested

Jolg42 (approved):
Looks great! 👍🏼  See comments

.github/scripts/test-project.sh
janpio: Yep, using Migrate here via `npx prisma` is generally a tradeoff of this approach. If that ever breaks, the whole tests are in trouble. Not sure how to solve that, as we removed the top level dependency for a reason.

platforms-serverless/azure-functions-linux/run.sh
janpio: I so love this feature. Thanks!

.github/scripts/test-project.sh
millsp: A few thoughts for solutions:
* Maybe we could `cd` into the directory before running `npx`.
* Or a *special* folder with a `package.json` where we run that `npx`.

PR #1883, opened 2021-07-04
chore: replace napi and fix matrix order

janpio:
- This replaces the last occurences of `napi` that are not in the preview feature or ENV var (which will both go away soon by itself).
- It also changes the matrix order so that the engines type is less important than the actual feature under test


janpio (commented):
Needs review from @williamluke4, but approvals from others are welcome of course 👍

.github/scripts/check-client-binaries.sh
```diff
@@ -13,7 +13,7 @@ project=$2
 skipped_projects=(
   prisma-dbml-generator                   # No generated Client, so only Client stub with no engine included
   prisma-json-schema-generator            # No generated Client, so only Client stub with no engine included
-  napi-preview-feature                    #
+  engine-types                   #
```
janpio:
```suggestion
  engine-types                            #
```

PR #1911, opened 2021-07-08
fix(vercel-with-redwood, firebase-functions): update redwood and firebase projects

janpio:


.github/workflows/test.yaml
```diff
@@ -652,10 +652,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: use node 12
+      - name: use node 14
```
janpio, line 655: This is causing the Firebase tests to fail right now - we either need to fix these as well, or split apart (fixing preferred)

millsp:
It's all fixed except that vercel-with-redwood with napi is failing (not flaky).

PR #1955, opened 2021-07-29
chore(check-*-binaries): Output size of files as well

janpio:
The `ls` output now also outputs the filesize, chmod etc.


Jolg42:
Should we merge?

williamluke4 (approved):

PR #1957, opened 2021-07-30
refactor: Update set -eux in all test files

### LINKED ISSUE #1787: Use `set -eux` in all .sh files in tests
-x makes sure that commands are output before executing, that helps debug things generally.

Unless there is a reason not to of course.

ShubhankarKG:
Changes only files that are named as 'test.sh'

Fixes #1787

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -eu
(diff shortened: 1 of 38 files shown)


janpio (approved):
Although for some reason this does not trigger our tests, should be fine.

aqrln (approved):

PR #1964, opened 2021-08-03
test(engines/engine-types): update snapshot (added LICENSE)

Jolg42:


Jolg42:
Looks some snapshots still need an update

PR #2002, opened 2021-08-16
fix(aws-graviton): rm node_modules before install

williamluke4:


Jolg42:
Might fix this flaky error https://github.com/prisma/e2e-tests/runs/3325607937#step:9:439
```


  ● Prisma › should use the correct engine files

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `Prisma should use the correct engine files 1`

    - Snapshot  - 0
    + Received  + 1

    @@ -2,7 +2,8 @@
        "index-browser.js",
        "index.d.ts",
        "index.js",
        "libquery_engine-linux-arm-openssl-1.0.x.so.node",
        "package.json",
    +   "query-engine-linux-arm-openssl-1.0.x",
        "schema.prisma",
      ]

      25 |     const files = fs.readdirSync(path.dirname(require.resolve('.prisma/client')))
      26 |     if (process.env.PRISMA_CLIENT_ENGINE_TYPE !== 'binary') {
    > 27 |       expect(files).toMatchInlineSnapshot(`
         |                     ^
      28 | Array [
      29 |   "index-browser.js",
      30 |   "index.d.ts",

```

Jolg42:
Thanks @williamluke4 !

janpio:
Hmm, should not _really_ make a difference on that one @Jolg42 as the folder name includes the run ID, which is either library or binary all the time, on all retries.

(Does this even really work btw without creating the folder first manually? Or where is it created?)

Jolg42:
For @williamluke4 ⬆️

PR #2024, opened 2021-08-27
chore: sync, use 2.30.1-dev.2

millsp:


.github/scripts/check-for-update.sh
```diff
@@ -33,18 +33,18 @@ echo ""
...
+# git config --global user.name "Prismo"
 
-git remote add github "git@github.com:$GITHUB_REPOSITORY.git"
-git fetch github "$branch"
-git reset --hard "github/$branch"
-git checkout "github/$branch"
+# git remote add github "git@github.com:$GITHUB_REPOSITORY.git"
+# git fetch github "$branch"
+# git reset --hard "github/$branch"
+# git checkout "github/$branch"
```
janpio, line 47: scripts should only run from `main`, so this should be safe to leave unchanged and active.

millsp: I did that to skip "prismo"-related stuff so that I could bump the versions. It gets committed by the script. It would be nice maybe to guard these by an env var (or abscence of it) so I don't have to comment it out in the future (Ah but that probably won't be a problem anymore when we switch to pnpm :smile:)

janpio: Ah, you used this script yourself to do the update? Because usually it is only triggered via the GH Actions `check-for-update` on schedule - so having it lying around in branches is usually fine.

millsp: Yep, for manual update. Is there a better way to do this?

Jolg42:
Closing because we don't want to merge, just look at the CI results here.

PR #2048, opened 2021-09-03
test: add migrate db seed test

### LINKED ISSUE #1806: Test seeding

Jolg42:
Based on and replaces https://github.com/prisma/e2e-tests/pull/1807

closes #1806

@janpio this adds 12 jobs 😨


Jolg42:
I think we should remove the enginetype from the matrix here, it doesn’t bring any value I can see (it's tested in other tests and that would reduce the number of jobs here)

millsp (approved):

migrate/db-seed-commonjs-pkg/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "prisma": "2.31.0-dev.46",
+    "ts-node": "10.2.1",
+    "jest": "26.6.3",
+    "prettier": "2.3.1"
+  },
+  "dependencies": {
+    "@prisma/client": "2.31.0-dev.46"
+  },
+  "scripts": {
+    "test": "jest --forceExit"
```
janpio: Why `forceExit`? I think we should Jest let us tell about stuff that stays open:
```suggestion
    "test": "jest"
```

migrate/db-seed-commonjs-pkg/prisma/seed.ts
```diff
@@ -0,0 +1,71 @@
+import { PrismaClient, Prisma } from '@prisma/client'
+
+const prisma = new PrismaClient()
+
+const userData: Prisma.UserCreateInput[] = [
```
janpio, line 5: Is this best practice how to define input for our create methods? 💡

migrate/db-seed-commonjs-pkg/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -eux
+
+yarn prisma db pull --print
```
janpio, line 5: confirmation that `db push` worked?

migrate/db-seed-commonjs-pkg/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -eux
+
+yarn install
+
+# generate not needed, it will be done by db push
+# yarn prisma generate
+yarn prisma db push --force-reset
```
janpio, line 9: if `force-reset` does not work for whatever reason, the test would possibly also succeed if the seeding fails, correct?

migrate/db-seed-esm-pkg/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "prisma": "2.31.0-dev.46",
+    "ts-node": "10.2.1",
+    "jest": "26.6.3",
+    "prettier": "2.3.1"
+  },
+  "dependencies": {
+    "@prisma/client": "2.31.0-dev.46"
+  },
+  "scripts": {
+    "test": "jest --forceExit"
```
janpio:
```suggestion
    "test": "jest"
```

janpio (approved):

migrate/db-seed-commonjs-pkg/test.sh
Jolg42: It's a copy paste from your script, so probably

migrate/db-seed-commonjs-pkg/prisma/seed.ts
Jolg42: If you want to extract the object outside the prisma call, yes

migrate/db-seed-commonjs-pkg/run.sh
Jolg42: Well here it's about preparing the db and I assume test.sh runs after with
```
#!/bin/sh

set -eux

yarn prisma db pull --print

yarn prisma db seed

yarn test
```

This should fail

PR #2085, opened 2021-09-22
fix: studio tests

sdnts:


core-features/studio/test.ts
```diff
@@ -34,10 +34,10 @@ describe('Studio', () => {
     let res = await fetch(`http://localhost:${STUDIO_PORT}`)
     expect(res.status).toBe(200)
 
-    res = await fetch(`http://localhost:${STUDIO_PORT}/databrowser.js`)
+    res = await fetch(`http://localhost:${STUDIO_PORT}/assets/index.js`)
     expect(res.status).toBe(200)
 
-    res = await fetch(`http://localhost:${STUDIO_PORT}/index.css`)
+    res = await fetch(`http://localhost:${STUDIO_PORT}/http/databrowser.js`)
```
sdnts, line 40: Location of assets changed, so these tests were failing (valid)

core-features/studio/test.sh
```diff
@@ -21,7 +21,7 @@ yarn prisma studio -p 5555 -b none &
 PRISMA_PID=$!
 
 sleep 3 # Studio takes some time to start up
-yarn jest
+yarn jest --ci --runInBand
```
sdnts, line 24: This is going to make it so tests run one at a time (more like Jest will not use workers, but same result).

Not sure how tests were passing earlier, tests with a common DB cannot run in parallel 🤔

Jolg42 (approved):
Thanks for the quick fix!

PR #2110, opened 2021-10-06
feat: init docker alpine tests

pantharshit00:
This PR introduces a new docker folder which shall container all of our docker related tests. 

Inside of that folder, we have a test for alpine linux for now. Mostly reusing artefacts from the pm2 test.


pantharshit00:
Actually, ready for merge. Tests failure are expected.

docker/alpine/schema.prisma
```diff
@@ -0,0 +1,14 @@
+generator client {
```
Jolg42: Why 2 schemas? one in docker/alpine/schema.prisma and one in docker/alpine/prisma/schema.prisma

Jolg42:
Both library and engine jobs are failing with 
```
Error: Unable to require(`/usr/src/app/node_modules/@prisma/engines/libquery_engine-linux-musl.so.node`)
```

I expected the engine job the fail with the binary engine 🤔

docker/alpine/schema.prisma
pantharshit00: Copy/paste from other e2e test. I deleted the extra one.

pantharshit00:
> Both library and engine jobs are failing with
> 
> ```
> Error: Unable to require(`/usr/src/app/node_modules/@prisma/engines/libquery_engine-linux-musl.so.node`)
> ```
> 
> I expected the engine job the fail with the binary engine 🤔


Good catch! Looks like I forgot to pass the engine type env variable to docker. 

Edit: Looks like CLI still tried to open the library file regardless of the environment variable.

Jolg42:
@pantharshit00 I think you'll need to use `PRISMA_CLI_QUERY_ENGINE_TYPE` (we use that in prisma/prisma)

janpio (approved):

PR #2141, opened 2021-10-23
test(dataproxy): vercel and cfws

### LINKED ISSUE #9596: CFW: Add E2E tests
- [x] Deploy to CFW
- [x] Deploy to Vercel Edge Functions

millsp:
fixes https://github.com/prisma/prisma/issues/9596


.github/workflows/test.yaml
```diff
@@ -722,6 +724,55 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - cloudflare-workers
+          - vercel-edge-functions
+        clientEngine: ['dataproxy']
+        os: [ubuntu-latest] #, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    concurrency: ${{ github.run_id }}-platforms-serverless-${{ matrix.platform }}
```
janpio: this should to be removed or adapted (`platforms-serverless` => `data-proxy`) for consistency

.github/workflows/test.yaml
```diff
@@ -722,6 +724,55 @@ jobs:
...
+    env:
+      START_TIME: ${{ needs.start-time.outputs.start-time }}
+      NPM_CONFIG_LOGLEVEL: error
+      NODE_ENV: development
+      NODE_MODULES_CACHE: false
+      NODE_VERBOSE: true
+      AES_256_PASS: ${{ secrets.AES_256_PASS }}
```
janpio, line 746: Can we give this secret a more descriptive name? Pass to what?

.github/workflows/test.yaml
```diff
@@ -722,6 +724,55 @@ jobs:
...
+      PRISMA_CLIENT_ENGINE_TYPE: dataproxy
```
janpio: Should this potentially be dynamic via `matrix.clientEngine`?

.github/workflows/test.yaml
millsp: We ran out of secrets, so I started a new thing https://github.com/prisma/e2e-tests/pull/2141/files#diff-73675af499015e989f5ef5d2b73578045fd72da0a3da6bbdc09593e1375ace8dR5 that decrypts and exports env vars via https://github.com/prisma/e2e-tests/pull/2141/files#diff-a952e0014885757e976dc08bf8f819f60447dd452ebe828bd5684868a53d155aR5. They are encrypted and decrypted via the `AES_256_PASS` https://github.com/prisma/e2e-tests/pull/2141/files#diff-5bc97016adbf2b476d3412f84c0f12c70dfd3e4d37d41cf0327a6875b78981d5R17. I inspired from their docs to work around this 100-secret limitation, [in this section](https://docs.github.com/en/actions/security-guides/encrypted-secrets#limits-for-secrets).

.github/workflows/test.yaml
millsp: I'd probably rename this test folder to something like `v8` or `dataproxy-v8`. In this case, I just want to test on the two very specific serverless platforms that are `cloudflare-workers` and `vercel-edge-functions`. I think that it would make sense to flip that switch on the other workflows? Before doing that, I'd like to have a way to init a dataproxy with our usual `DATABASE_URL`.

.github/workflows/test.yaml
janpio: Ah that makes sense. 

I think we can actually delete a _lot_ of the existing secrets. Will make a TODO for myself to clean up the list some time soon.

.github/workflows/test.yaml
janpio: I don't think I understand you answer. But if it works, this is perfectly fine of course.

millsp: Yep, sorry. The confusion came from my side! I'll exclude `binary` and `library` engines for these two tests and put them back into `platforms-serverless`. Thanks!

.github/scripts/check-engines-cli.sh
```diff
@@ -99,7 +101,7 @@ echo "---"
 
 # TODO Add test that makes sure not _wrong_ files are present as well
 # Example: `community-generators (napi, prisma-dbml-generator)` has correct node_modules/prisma/libquery_engine-debian-openssl-1.1.x.so.node, but wrong node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x (also `community-generators (napi, prisma-json-schema-generator)`)
-if [ -f "$qe_location" ]  || [ -f "$qe_location2" ] ; then
+if [ -f "$qe_location" ] || [ -f "$qe_location2" ] || [ $CLI_QUERY_ENGINE_TYPE == "dataproxy" ] ; then
```
janpio: Should we maybe do a special case here that output "no query engine required as type dataproxy"? I think right now there are actually some files present, that we might not want in the future etc.

.github/scripts/check-engines-client.sh
```diff
@@ -99,7 +101,7 @@ fi
 echo "--- ls -lh node_modules/.prisma/client/ ---"
 ls -lh node_modules/.prisma/client/
 echo "---"
-if [ -f "$qe_location" ] ; then
+if [ -f "$qe_location" ] || [ $CLIENT_ENGINE_TYPE == "dataproxy" ] ; then
```
janpio: same as https://github.com/prisma/e2e-tests/pull/2141/files#r744629673

utils/crypto/envVars.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const envVars = {
+    CF_DATA_PROXY_URL: "U2FsdGVkX18aWMCedtp1nKGy8SGsipaJIw5NqyI4ZNQaKmB3f2jnmerKKp5JpAHVLHiQ+8oO3MU7DYhQeaR4RYZaErFgl7Ox1EZ4UBpPFhN10kxZvd5sOOnxhd+WQ4My9l+aOhmF+iz6m3E+k3YxldUqtp/5ej4Xr0NW/toml2RVI+FItfA2GC4Df9PvyfnP",
+    CF_ACCOUNT_ID: "U2FsdGVkX1/kgKmF6Nc3rIIkben8az6boItdyzPJFj5vEPoxnWtfpyjEO8E6voHLB4Zl/+kddgTA14uboA+40A==",
+    CF_API_TOKEN: "U2FsdGVkX18fx8PpClVvhLV9TuBGm//20tdPUYRowj/hBUw+r7nECbfZ7DBPKgI9lelSXXKOXD4FFNg6FPpURA==",
+
+    VERCEL_DATA_PROXY_URL: "U2FsdGVkX18aWMCedtp1nKGy8SGsipaJIw5NqyI4ZNQaKmB3f2jnmerKKp5JpAHVLHiQ+8oO3MU7DYhQeaR4RYZaErFgl7Ox1EZ4UBpPFhN10kxZvd5sOOnxhd+WQ4My9l+aOhmF+iz6m3E+k3YxldUqtp/5ej4Xr0NW/toml2RVI+FItfA2GC4Df9PvyfnP",
+    VERCEL_EDGE_FUNCTIONS_ORG_ID: "U2FsdGVkX1+OynKKazwnZh8yRbrGoapHZmyTDNe2ryNIFCGS/x8ALH8xROfzCdz+",
+    VERCEL_EDGE_FUNCTIONS_PROJECT_ID: "U2FsdGVkX1/rvuHFq0MrRcXU3LmvnzZQp8A9/WeoqYg8IllvAJhujSnQOVfSvbqvfBGzk7t/ewBV/uXyQ99Jwg=="
+}
```
janpio: separate file maybe?

utils/crypto/envVars.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# equivalent of a __dirname in node.js
+DIR=$(dirname "${BASH_SOURCE[0]}")
+
+# makes sure which node_modules to use
+MOD_DIR="$DIR/../../node_modules"
+
+# workaround for an error with sourcing
+TS_NODE=$(yarn --modules-folder $MOD_DIR bin ts-node)
```
janpio: OT: why is this needed?

workers/cloudflare-workers/prepare.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+export PRISMA_TELEMETRY_INFORMATION='e2e-tests data-proxy cloudflare-workers build'
```
janpio:
```suggestion
export PRISMA_TELEMETRY_INFORMATION='e2e-tests workers cloudflare-workers build'
```

workers/vercel-edge-functions/prepare.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+export PRISMA_TELEMETRY_INFORMATION='e2e-tests data-proxy vercel-edge-functions build'
```
janpio:
```suggestion
export PRISMA_TELEMETRY_INFORMATION='e2e-tests workers vercel-edge-functions build'
```

workers/cloudflare-workers/index.test.ts
```diff
@@ -0,0 +1,69 @@
+/// <reference types="@types/jest" />
+
+import fetch from 'node-fetch'
+
+describe('use data proxy', () => {
+  test('fetch response', async () => {
+    jest.setTimeout(30000)
+
+    const response = await fetch(process.env.DEPLOYMENT_URL!)
```
janpio: to make it more explicit what this is:
```suggestion
    const cloudflareWorkersDeploymentUrl = process.env.DEPLOYMENT_URL!
    const response = await fetch(cloudflareWorkersDeploymentUrl )
```

.github/workflows/test.yaml
```diff
@@ -722,6 +724,54 @@ jobs:
...
+  workers:
+    needs: start-time
+    timeout-minutes: 60 # can take longer if platforms are down, so better protect
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - cloudflare-workers
+          - vercel-edge-functions
+        clientEngine: ['dataproxy']
```
```suggestion
        clientEngine: ['dataproxy'] # only valid value here
```

workers/vercel-edge-functions/README.md
```diff
@@ -0,0 +1,52 @@
+# JSON Response Example
```
janpio: This file should probably be replaced with our usual minimal README for test projects.

workers/vercel-edge-functions/index.test.ts
```diff
@@ -0,0 +1,69 @@
+/// <reference types="@types/jest" />
+
+import fetch from 'node-fetch'
+
+describe('use data proxy', () => {
+  test('fetch response', async () => {
+    jest.setTimeout(30000)
+
+    const response = await fetch(process.env.DEPLOYMENT_URL!)
```
janpio:
```suggestion
    const vercelEdgeFunctionsDeployment = process.env.DEPLOYMENT_URL!
    const response = await fetch(vercelEdgeFunctionsDeployment)
```

workers/vercel-edge-functions/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "license": "MIT",
+  "scripts": {
+    "test": "jest",
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start"
+  },
+  "dependencies": {
+    "@prisma/client": "3.5.0-integration-data-proxy-client-improvements.6",
+    "next": "11.1.3-canary.104",
```
janpio: is there a more stable version? Renovate might have problems keeping that up to date.

workers/vercel-edge-functions/prepare.sh
```diff
@@ -0,0 +1,12 @@
...
+
+source ../../utils/crypto/envVars.sh VERCEL_DATA_PROXY_URL VERCEL_EDGE_FUNCTIONS_PROJECT_ID VERCEL_EDGE_FUNCTIONS_ORG_ID
+
+# we set the vercel project & organization environment variables
+export VERCEL_PROJECT_ID=$VERCEL_EDGE_FUNCTIONS_PROJECT_ID
+export VERCEL_ORG_ID=$VERCEL_EDGE_FUNCTIONS_ORG_ID
```
janpio, line 9: Is this a different org than our other proejects?

janpio (approved):

utils/crypto/envVars.sh
millsp: Since this file is sourced inside of another project, `yarn ts-node` might not be available in that project's `node_modules`. So we make sure that we use the one that is in the root `node_modules`.

workers/vercel-edge-functions/README.md
millsp: Totally forgot, thanks!

workers/vercel-edge-functions/prepare.sh
millsp: Nope, should I use the env var instead?

.github/scripts/check-engines-cli.sh
millsp: Will do.

workers/vercel-edge-functions/package.json
millsp:
```suggestion
    "next": "12.0.2",
```

workers/vercel-edge-functions/prepare.sh
janpio: Not super important - we probably do that with some of the other Vercel projects then this follows just that pattern. If we do _not_ do that, I think it would be a small but nice simplification to also reuse the same env var.

PR #2196, opened 2021-11-26
fix(serverless-framework-lambda): Make retries work by overwriting credentials config

janpio:
Avoids:

```
$ /home/runner/work/e2e-tests/e2e-tests/platforms-serverless/serverless-framework-lambda/node_modules/.bin/serverless config credentials --provider aws --key *** --secret ***
Serverless: Setting up AWS...
 
 Serverless Error ----------------------------------------
 
  Profile "default" is already configured in ~/.aws/credentials. Use the overwrite flag ("-o" or "--overwrite") to force the update.
```


janpio:
Test run here "unfortuantely" did not require a rerun, so let's just merge and see if it helps to make that test less flaky.

PR #2213, opened 2021-12-03
feat: Run additional CLI commands for the `pkg` CLI binaries

janpio:


binaries/pkg/test.sh
```diff
@@ -25,3 +25,11 @@ esac
 yarn pkg node_modules/prisma -t node12-$os
 
 ./$filename --version
+
+./$filename
+
+./$filename init --datasource-provider sqlite
```
janpio: This fails because of https://github.com/prisma/prisma/blob/bbdf1c23653a77b0b5bf7d62efd243dcebea018b/packages/cli/src/Init.ts#L159-L166 right now :/

janpio: Potentially fixed via https://github.com/prisma/prisma/pull/10568

janpio: Worked around by changing the binary name if necessary.

PR #2250, opened 2021-12-24
fix(scripts): check the correct variable in check-engines-cli.sh

aqrln:
Fixes

```
../../.github/scripts/check-engines-cli.sh: line 104: [: ==: unary operator expected
```

### DIFF
--- a/.github/scripts/check-engines-cli.sh
+++ b/.github/scripts/check-engines-cli.sh
@@ -1,19 +1,29 @@
 #!/bin/bash
 
+set -e
+
 echo "-------------- Checking CLI/Engines QE Engine --------------"
 
 DIR=$1
 PROJECT=$2
 
-DEFAULT_CLI_QUERY_ENGINE_TYPE='library'
+DEFAULT_QUERY_ENGINE_TYPE='library'
 
 # Check to see if the env var "PRISMA_CLI_QUERY_ENGINE_TYPE" is set if not then using the default
 if [ -z "$PRISMA_CLI_QUERY_ENGINE_TYPE" ]; then
-  echo "Using default cli qe: $DEFAULT_CLI_QUERY_ENGINE_TYPE"
-  CLI_QUERY_ENGINE_TYPE=$DEFAULT_CLI_QUERY_ENGINE_TYPE
+  echo "Using default cli qe: $DEFAULT_QUERY_ENGINE_TYPE"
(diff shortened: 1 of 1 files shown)


aqrln:
Adding `set -e` uncovered another issue

```
--- yarn prisma -v ---
Unknown Syntax Error: Command not found; did you mean one of:

  0. yarn cache clean [--mirror] [--all]
  1. yarn cache clean [--mirror] [--all]
  2. ....
```

This is something to be fixed separately, I'll remove `set -e` back for now.

aqrln:
https://github.com/prisma/e2e-tests/issues/2251

.github/scripts/check-engines-cli.sh
```diff
@@ -5,17 +5,25 @@ echo "-------------- Checking CLI/Engines QE Engine --------------"
 DIR=$1
 PROJECT=$2
 
-DEFAULT_CLI_QUERY_ENGINE_TYPE='library'
+DEFAULT_QUERY_ENGINE_TYPE='library'
 
 # Check to see if the env var "PRISMA_CLI_QUERY_ENGINE_TYPE" is set if not then using the default
```
Jolg42:
```suggestion
# Check to see if the env var "DEFAULT_QUERY_ENGINE_TYPE" is set if not then using the default
```

janpio: But the code checks for `PRISMA_CLI_QUERY_ENGINE_TYPE`.

Jolg42:
It's confusing to me that there is "Client" logic in this file named `check-engines-cli.sh` and that we also have a file called `check-engines-client.sh` where it would make more sense to me https://github.com/prisma/e2e-tests/blob/aqrln/fix-check-engines-cli-script/.github/scripts/check-engines-client.sh

aqrln:
Also it seems that this logic is actually present in `check-engines-client.sh`. ~~The dataproxy check should probably just be removed from `check-engines-cli.sh`.~~

aqrln:
Oh, I see. Given these lines:

https://github.com/prisma/e2e-tests/blob/c2c421098204b12191d95db9900716136bc8e6e1/.github/scripts/check-engines-cli.sh#L94-L95

It looks like it probably was a typo and `$CLI_QUERY_ENGINE_TYPE` was meant to be checked instead of `$CLIENT_ENGINE_TYPE`.

Jolg42 (approved):
Nice!

PR #2269, opened 2022-01-10
fix(utils): disable cmd printing in setEnv

millsp:


Jolg42 (approved):
Nice!

Jolg42:
Now looks like 
https://github.com/prisma/e2e-tests/runs/4763250446?check_suite_focus=true#step:6:90
```
++ source ../../utils/crypto/setEnv.sh CF_ACCOUNT_ID CF_API_TOKEN CF_DATA_PROXY_URL
+++ [[ ehuxB == *x* ]]
+++ set +x
```
and
https://github.com/prisma/e2e-tests/runs/4763250832?check_suite_focus=true#step:6:85
```
 ++ source ../../utils/crypto/setEnv.sh VERCEL_DATA_PROXY_URL VERCEL_EDGE_FUNCTIONS_PROJECT_ID VERCEL_EDGE_FUNCTIONS_ORG_ID
+++ [[ ehuxB == *x* ]]
+++ set +x
```

Jolg42:
Looks good to merge, the failures are unrelated

utils/crypto/setEnv.sh
```diff
@@ -11,3 +18,8 @@ TS_NODE=$(yarn --modules-folder $MOD_DIR bin ts-node)
 
 # execute what ts-node has printed out
 eval $($TS_NODE $DIR/setEnv.ts "$@")
+
+# reactivate -x flag for cmd printing
+if [[ $HAS_X_FLAG == true ]]; then
```
aqrln, line 23: This would fail with unbound variable error if `x` flag wasn't set, since the upstream scripts that source this also have `set -u` flag set.

Either `HAS_X_FLAG` needs to be assigned to something before the first `if` statement, or this needs to be changed to

```suggestion
if [[ "${HAS_X_FLAG-}" == true ]]; then
```

PR #2270, opened 2022-01-10
fix(utils): prevent possible unbound variable error

aqrln:
See https://github.com/prisma/e2e-tests/pull/2269/files#r781385908


Jolg42 (approved):
Thanks!

PR #2285, opened 2022-01-11
fix(bundlers): bundle externals

### LINKED ISSUE #2171: Bundler tests don't bundle the prisma client
That is an issue because it does not reflect what our users do. As far is I know, prisma is marked as external for `webpack`.

The next steps:
* Reproduce https://github.com/prisma/prisma/issues/10058 (I have it on my local)
* This issue happens via `@prisma/client@3.4.2`, so the idea is to make sure that it fails with that.
* Since this issue also has already been fixed, the e2e test should not fail with the latest prisma.

millsp:
Fixes https://github.com/prisma/e2e-tests/issues/2171

### DIFF
--- a/bundlers/webpack/package.json
+++ b/bundlers/webpack/package.json
@@ -7,8 +7,7 @@
     "jest": "27.4.7",
     "prisma": "3.8.0-dev.62",
     "webpack": "5.65.0",
-    "webpack-cli": "4.9.1",
-    "webpack-node-externals": "3.0.0"
+    "webpack-cli": "4.9.1"
   },
   "dependencies": {
     "@prisma/client": "3.8.0-dev.62"
--- a/bundlers/webpack/webpack.config.js
+++ b/bundlers/webpack/webpack.config.js
@@ -1,6 +1,7 @@
-const nodeExternals = require('webpack-node-externals')
-
 module.exports = {
   target: 'node',
-  externals: [nodeExternals()],
+  externals: {
+    "_http_common": "commonjs2 _http_common",
+  }
 }
+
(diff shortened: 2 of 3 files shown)


aqrln (approved):

Jolg42:
You might wat to check this/nderstand it
https://github.com/prisma/e2e-tests/runs/4775213666?check_suite_focus=true#step:6:131
```
WARNING in ./node_modules/@prisma/client/runtime/index.js 14993:18-45
Module not found: Error: Can't resolve 'encoding' in '/home/runner/work/e2e-tests/e2e-tests/bundlers/webpack/node_modules/@prisma/client/runtime'
 @ ./node_modules/.prisma/client/index.js 17:4-43 134:29-68 59:21-54
 @ ./node_modules/@prisma/client/index.js 1:15-46
 @ ./src/index.js 1:0-45 3:19-31

WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value.
Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: webpack.js.org/configuration/mode
```

millsp:
Good point. `encoding` suffers from the same issues as `_http_common`, though they are from different reasons. I remember there were two solutions for that one:
- Upgrade the dependency where the requirement comes from (i think it was `fetch`, can you can confirm that @aqrln?).
- This dependency is totally optional, since we don't use it and did not mark it as peerDep then fill it out (empty module).
- Mark it as external like I just did but totally unituitive from a user perspective - so to be fixed in the future.

aqrln:
@millsp yeah, it's `node-fetch`, as per @Jolg42's comment here: https://github.com/prisma/prisma/pull/10978#issuecomment-1009162476

PR #2286, opened 2022-01-11
feat: detect jobs to run

### LINKED ISSUE #1746: PRs: Only run tests that have changes
A PR like https://github.com/prisma/e2e-tests/pull/1653 or https://github.com/prisma/e2e-tests/pull/1722 would really benefit from only running the tests where files actually change - no way other tests should be influenced by that.

Jolg42:
Closes https://github.com/prisma/e2e-tests/issues/1746

Based on https://github.com/prisma/prisma/pull/6612 and improved


.github/workflows/test.yaml
```diff
@@ -88,7 +115,8 @@ jobs:
   ### real jobs after this
 
   process-managers:
-    needs: start-time
+    needs: [start-time, detect_jobs_to_run]
+    if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-process-managers-') }}
```
```suggestion
    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['process-managers'] == true
```

.github/workflows/test.yaml
```diff
@@ -28,6 +28,33 @@ jobs:
...
         run: echo "::set-output name=start-time::$(date)"
 
+  # From https://github.com/marketplace/actions/skip-duplicate-actions
+  # This action cleans up previously running instances of a workflow on the same branch.
+  # This accomplishes the task of automatically cancelling CI runs on pushes to the same branch,
+  # which is a common feature in most CI systems but currently not possible with GitHub actions.
+  cleanup-runs:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/main' && !contains(github.actor, 'renovate')"
```
```suggestion
    if: "!startsWith(github.ref, 'refs/tags/') && !contains(github.actor, 'renovate') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/integration' && github.ref != 'refs/heads/patch-dev' && github.ref != 'refs/heads/latest')"
```

.github/workflows/detect-jobs-to-run.js
```diff
@@ -0,0 +1,113 @@
...
+    'frameworks',
+    'platforms',
+    'paltforms-serverless',
+    'workers',
+    'bundlers',
+    'libraries',
+    'databases',
+    'databases-macos',
+    'test-runners',
+  ]
```
janpio: If in any way possible we should avoid duplicating this list. Maybe just filter out the folders that are _not_ tests? That should be more static.

.github/workflows/test.yaml
```diff
@@ -88,7 +115,8 @@ jobs:
   ### real jobs after this
 
   process-managers:
-    needs: start-time
+    needs: [start-time, detect_jobs_to_run]
+    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['process-managers'] == true
```
janpio: `process-managers` should be available as an ENV var here, making this copy pastable to all jobs without adaptations (which we would mess up sooner or later)

I _think_ it is `${{ github.job }}` (also used further down in "Set test specific DATABASE_URL" step)

Jolg42: Good point!

.github/workflows/detect-jobs-to-run.js
Jolg42: Yep maybe doing the same as in 
https://github.com/prisma/e2e-tests/blob/60f3b664ca7b66e63904c266528bf3e2a3f3ef58/.github/scripts/check-folders.js

```
  const jobKeys = Object.keys(optionalTestYaml['jobs'])
    .concat(Object.keys(testYaml['jobs']))
    .filter...
```

.github/workflows/test.yaml
```diff
@@ -28,6 +28,33 @@ jobs:
...
         run: echo "::set-output name=start-time::$(date)"
 
+  # From https://github.com/marketplace/actions/skip-duplicate-actions
+  # This action cleans up previously running instances of a workflow on the same branch.
+  # This accomplishes the task of automatically cancelling CI runs on pushes to the same branch,
+  # which is a common feature in most CI systems but currently not possible with GitHub actions.
+  cleanup-runs:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    if: "!startsWith(github.ref, 'refs/tags/') && !contains(github.actor, 'renovate') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/integration' && github.ref != 'refs/heads/patch-dev' && github.ref != 'refs/heads/latest'"
```
millsp: Why not allow this on renovate?

janpio: Is there a way to skip this? On branches one might want to try a few things in quick succession by committing the variants and then look at all the CI results.

.github/workflows/test.yaml
Jolg42: It didn't work :meh:

.github/workflows/test.yaml
Jolg42: You mean :trollface:  '!contains(github.actor, 'janpio')`? 

Did a copy paste, not sure we disabled it on renovate, it would be useful to have it, especially for renovate 😄

.github/workflows/test.yaml
janpio: Why not? The syntax to use it in there is probably different, maybe just `if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)[github.job] == true` - was that one of the variants you tried? (I never get these variable substitutions right on the first try)

.github/workflows/test.yaml
```diff
@@ -28,6 +28,38 @@ jobs:
...
+    steps:
+      - uses: fkirc/skip-duplicate-actions@v3.4.1
```
janpio: Can/Should we maybe put this into a separate PR so we can try/merge each in isolation and see what happens?

.github/workflows/test.yaml
```diff
@@ -570,9 +608,10 @@ jobs:
         run: bash .github/slack/notify-failure.sh ${{ github.job }} ${{ matrix.framework }}
 
   platforms:
-    needs: start-time
-    timeout-minutes: 60
+    needs: [start-time, detect_jobs_to_run]
+    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['platfroms'] == true }}
```
```suggestion
    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['platforms'] == true }}
```

.github/workflows/test.yaml
```diff
@@ -634,7 +673,9 @@ jobs:
         run: bash .github/slack/notify-failure.sh ${{ github.job }} ${{ matrix.platform }}
 
   platforms-serverless:
-    needs: start-time
+    needs: [start-time, detect_jobs_to_run]
+    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['platfroms-serverless'] == true }}
```
```suggestion
    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['platforms-serverless'] == true }}
```

.github/workflows/detect-jobs-to-run.js
```diff
@@ -0,0 +1,116 @@
...
+
+  for (const directoryName of testDirectories) {
+    // If changes are located only in one of the directories
+    if (
+      filesChanged.every((fileChanged) => fileChanged.startsWith(directoryName))
+    ) {
+      jobsToRun[directoryName] = true
+    } else {
+      jobsToRun[directoryName] = false
+      fallbackToAll = true
```
janpio: Whoooh what does that do here?

.github/workflows/detect-jobs-to-run.js
```diff
@@ -0,0 +1,116 @@
...
+  console.debug('fallbackToAll:', fallbackToAll)
+  console.debug('jobsToRun:', jobsToRun)
+
+  if (fallbackToAll) {
+    console.debug('Fallback! All directories must be tested.')
+    // creates an object with all values set to true
+    const fallbackRunAll = testDirectories.reduce(
+      (acc, curr) => ((acc[curr] = true), acc),
+      {},
+    )
```
janpio: Thanks for this comment - I would never understand this code :p

.github/workflows/detect-jobs-to-run.js
```diff
@@ -0,0 +1,116 @@
...
+    console.debug('fallbackRunAll:', fallbackRunAll)
+    console.log('::set-output name=jobs::' + JSON.stringify(fallbackRunAll))
+  } else {
+    console.debug('Only one directory will be tested')
```
janpio: Could this not also impact multiple directories at the same time?
```suggestion
    console.debug('Only the following directory/ies will be tested')
```

janpio (approved):

.github/workflows/test.yaml
Jolg42: it's now commented and not running

.github/workflows/test.yaml
Jolg42: I tried this and it didn't work [`385377c` (#2286)](https://github.com/prisma/e2e-tests/pull/2286/commits/385377cb114d0b953c7f77a56ab433704fa098f6)
```
if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['process-managers'] == true }}
```

.github/workflows/test.yaml
```diff
@@ -28,6 +28,38 @@ jobs:
...
         run: echo "::set-output name=start-time::$(date)"
 
+  # # From https://github.com/marketplace/actions/skip-duplicate-actions
+  # # This action cleans up previously running instances of a workflow on the same branch.
+  # # This accomplishes the task of automatically cancelling CI runs on pushes to the same branch,
+  # # which is a common feature in most CI systems but currently not possible with GitHub actions.
+  # cleanup-runs:
+  #   continue-on-error: true
+  #   runs-on: ubuntu-latest
+  #   if: "!startsWith(github.ref, 'refs/tags/') && !contains(github.actor, 'renovate') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/integration' && github.ref != 'refs/heads/patch-dev' && github.ref != 'refs/heads/latest'"
```
millsp:
```suggestion
  #   if: (!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/integration' && github.ref != 'refs/heads/patch-dev' && github.ref != 'refs/heads/latest') || contains(github.actor, 'renovate') 
```

millsp: Maybe GH was evaluating this as always thruthy because of the quotes - or I'm not aware that this is a "valid" conditional syntax. I see a lot of value in cancelling running workflows, so I'd like it to work :)

millsp (changes requested):

.github/workflows/detect-jobs-to-run.test.js
```diff
@@ -1,9 +1,13 @@
 const { detectJobsTorun } = require('./detect-jobs-to-run')
 // @ts-check
 
+//
+// This is running automatically with the `lint test suite` job
```
janpio: Renamed to "test-and-lint" then maybe?

janpio (approved):

.github/workflows/test.yaml
Jolg42: Thanks not sure what happened here I'll figure out in another PR later
```suggestion
  # contains(github.actor, 'renovate') 
  #   if: (!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/integration' && github.ref != 'refs/heads/patch-dev' && github.ref != 'refs/heads/latest')
```

Jolg42:
First run on dev branch as expected https://github.com/prisma/e2e-tests/runs/4805232450?check_suite_focus=true
With fallback to all tests as it's the dev branch

Jolg42:
and here it only runs platforms when I only edit platforms 🎊 
https://github.com/prisma/e2e-tests/runs/4805583055?check_suite_focus=true

millsp:
Awesome! Thank you :partying_face: :100:

PR #2298, opened 2022-01-13
test: temporary disable serverless framework test

Jolg42:
It will make Renovate automerge work again

I will create the revert PR right away with Slack reminder after we merge this so we don't forget to reactivate once Dom unblocks the AWS policy


janpio:
Test failure of detect-jobs-to-run is not a problem?

Jolg42:
No it was because I created that branch before fixing the config

PR #2344, opened 2022-01-21
test(engine-types): increase timeout to 300s for flakyness

Jolg42:
The small investigation I did only revealed so far that we were often close to timeout in some runs but not sure why yet, this should avoid the timeout


janpio (approved):
Meh, but ok.

Jolg42:
Yeah 😒

PR #2379, opened 2022-01-31
test(dataproxy): node.js

millsp:
This should fail until the https://github.com/prisma/prisma/pull/11529 fix is merged.


Jolg42:
Indeed! https://github.com/prisma/e2e-tests/runs/5012588398?check_suite_focus=true#step:6:485
```
FAIL ./index.test.ts
  use data proxy
    ✕ fetch response (236 ms)

  ● use data proxy › fetch response


    Invalid `prisma.user.findFirst()` invocation:


      Invalid value "undefined" for header "traceparent"

      at cb (node_modules/@prisma/client/runtime/proxy.js:94:56)
          at async Promise.all (index 0)

```

dataproxy/cloudflare-workers/src/index.ts
```diff
@@ -3,7 +3,13 @@ import { PrismaClient } from '@prisma/client'
...
 
 async function getUsers() {
-  const data = await prisma.user.findMany()
+  // we use a transaction as it has caused issues with tracing
+  // `traceparent: undefined` was being passed to the engine
+  // but we ultimately want to test that queries are sent well
+  const data = await prisma.$transaction([
+    prisma.user.findFirst(),
+    prisma.user.findMany()
+  ])
```
Jolg42, line 12: But transaction was not supported on dataproxy as far as I know? Is it supported now?

millsp: Only regular transactions are supported, it's interactive transactions that aren't (yet).

Jolg42: Right! thanks

Jolg42 (approved):
Nice!

PR #2433, opened 2022-02-11
fix(vercel): Update org/team ID after account move

janpio:
This fixes the Vercel env vars after the project move to a new Vercel account.


millsp (approved):
Thank you!

PR #2452, opened 2022-02-16
feat: MongoDB at Azure CosmosDB

janpio:
Test for https://docs.microsoft.com/en-us/azure/cosmos-db/introduction


databases/mongodb-azure-cosmosdb/script.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const objectId = '6d795f757365725f69643030'
+
+  await client.user.create({
+    data: {
+      id: objectId,
+      email: 'alice@prisma.io',
+      name: 'Alice',
+    },
+  })
```
Jolg42, line 17: Looks like this create has concurrency issue (binary / library test running at the same time)
https://github.com/prisma/e2e-tests/runs/5219187664?check_suite_focus=true#step:6:153

Though it works in the end

janpio: Isn't that the same with all our database testing scripts? (deleteMany of `binary` can also just remove these creates of `library` again)

Jolg42: It's not amazing but it works with the retries

PR #2694, opened 2022-04-22
feat(framework/sveltekit): add sveltekit

janpio:


janpio:
Now used in https://github.com/prisma/ecosystem-tests/pull/2695

Issue #2728, opened 2022-04-30
Make Vercel jobs wait for at least the time the CLI returns on API rate limit events


janpio:
Our Vercel jobs sometimes hit API rate limits. The CLI responses include some information how long we should wait until we try again, but we are not using that information in our retry logic right now and instead retry pretty much instantly. This of course makes he problem worse.

```
Error! You have been creating deployments at a very fast pace. Please retry in 38 seconds.
```

Potential approach: Get strerr output in a file, parse it in `finally.sh` with a regex and then wait for x seconds before letting the script run the next retry.

Jolg42:
We do not have this issue anymore since our rate limit was uplifted.

janpio:
The problem still exists fundamentally - if we hit any API limits, the script will try again instantly.

Jolg42:
Sure, we do not need to do anything here, though. Since we do not see this issue anymore.

janpio:
🤞

PR #2729, opened 2022-05-02
Split out Vercel jobs to be able to remove retry

janpio:
The Vercel jobs are currently often failing because of API limits being hit.
This is made worse by our automated retries which do not wait between jobs.
Until https://github.com/prisma/ecosystem-tests/issues/2728 makes us wait properly between attempts, this PR disables at least the retries to not make the situation worse.


janpio:
- The failing tests are unrelated and fixed via https://github.com/prisma/ecosystem-tests/pull/2731
- This PR will go into conflicted state on each merge to `dev`, so I suggest you review/approve and I merge `dev` once before merging.

.github/scripts/check-folders.js
```diff
@@ -4,63 +4,81 @@ const yaml = require('yaml')
 const fs = require('fs')
 const path = require('path')
```
janpio, line 6: Small cleanup in this file to make it easier to understand, improvements on the comments welcome.

.github/workflows/test.yaml
```diff
@@ -778,6 +768,70 @@ jobs:
...
+          replace=${{ github.job }}_${{ matrix.platform }}_${{ matrix.os }}
+          replaced=${string/$search/$replace}
+          echo "DATABASE_URL=$replaced" >> $GITHUB_ENV
+
+      - name: test ${{ matrix.platform }} - ${{ matrix.clientEngine }}
+        id: run-test
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 60
+          max_attempts: 1
```
janpio, line 786: This is the important difference here that will leads to making the Vercel API situation less bad.

.github/scripts/check-folders.js
```diff
@@ -70,6 +88,7 @@ async function main() {
       return acc.concat([...folders])
     }, [])
 
+  // TODO ??
```
Jolg42: It's transforming the shape to return something that can be easy to use from GitHub Actions if conditionals

.github/scripts/check-folders.js
```diff
@@ -4,63 +4,81 @@ const yaml = require('yaml')
...
-        'start-time', // Not a test but a job that fills an env var with the job start time
-        'report-to-slack-success', // Not a test but a job that posts to slack
-        'report-to-slack-failure', // Not a test but a job that posts to slack
-        'detect_jobs_to_run', // Not a test but a job that decides which tests should run
-        'cleanup-runs', // Not a test but a job that cancels previous runs
-      ]
       return !jobsToIgnore.includes(key)
     })
+  
+  // TODO ??
```
Jolg42: Create an array of job names, from reading the `jobs` keys of the .yml file

Example of an item should look like `platforms-serverless`, which is also matching the directory name

.github/scripts/check-folders.js
Jolg42: Looks like 
```
Object {
  "binaries": true,
  "bundlers": true,
  "core-features": true,
  "databases": true,
  "databases-macos": true,
  "dataproxy": true,
  "docker": true,
  "engines": true,
  "frameworks": true,
  "libraries": true,
  "migrate": true,
  "node": true,
  "os": true,
  "packagers": true,
  "platforms": true,
  "platforms-serverless": true,
  "platforms-serverless-vercel": true,
  "process-managers": true,
  "test-runners": true,
}
```
See https://github.com/prisma/ecosystem-tests/blob/dev/.github/workflows/detect-jobs-to-run.test.js

Jolg42 (approved):
That's a very nice improvement!

Jolg42:
@janpio Looks like you will need to fix the conflicts and merge right away

janpio:
Will do after applying your suggestions on the comments.

.github/scripts/check-folders.js
janpio:
```suggestion
  // Transform the shape to return something that can be easy to use from GitHub Actions if conditionals
```

.github/scripts/check-folders.js
janpio:
```suggestion
  // Create an array of job names, from reading the `jobs` keys of the .yml file
```

PR #2755, opened 2022-05-06
chore: add code owners

millsp:


janpio (approved):

Jolg42:
People, get ready for notifications 😅 

We could scope down a bit in the future and have the "clearly" client tests owned only by client people here, for example.

PR #2765, opened 2022-05-09
netlify-cli: remove binaryTarget

janpio:


Jolg42:
[Note from Slack](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1652166728331939?thread_ts=1652123050.768669&cid=C02FNFLDUS3)

What I could verify is that a deploy when using netlify deploy on commit (needs to first setup git repo with the project) works
Build image is “20.04.4 LTS (Focal Fossa) - OpenSSL 1.1.1f 31 Mar 2020” but this is unrelated because in the end it’s going to AWS Lambda, and there we need the rhel-openssl-1.0.x “binaryTarget”
During build on netlify image NETLIFY env var is set which activates the netlify "hack" https://github.com/prisma/prisma/blob/main/packages/sdk/src/get-generators/utils/getBinaryPathsByVersion.ts#L35-L40

And that using netlify deploy  CLI doesn’t work, why?
Using the netlify CLI happens locally
Locally i’m on macOS for example, and NETLIFY  env var is not set
When calling netlify deploy it doesn’t trigger a build on their CI, it just creates a build locally and uploads it to their server and they deploy that, which is missing the “binaryTarget” for AWS

janpio:
Yep, confirmed exactly that here in the end by the failure. Good error message as well.

PR #2779, opened 2022-05-12
test: Add test for browser bundles at custom path

SevInf:
Reference: prisma/prisma#13291


bundlers/webpack-browser-custom-output/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "main": "index.js",
+  "license": "MIT",
+  "devDependencies": {
+    "jest": "27.5.1",
+    "prisma": "3.15.0-integration-12296-custom-path-pkg.1",
+    "webpack": "5.72.1",
+    "webpack-cli": "4.9.2"
+  },
+  "dependencies": {
+    "@prisma/client": "3.15.0-integration-12296-custom-path-pkg.1"
```
janpio: I guess this test woudl be failing right now with the normal `dev` version of Prisma Client?

bundlers/webpack-browser-custom-output/index.test.js
```diff
@@ -0,0 +1,7 @@
+const { Role } = require('./dist/prismaTest')
+
+test('correctly generates browser-compatible bundle for prisma at custom path', () => {
+  expect(Role).toBeDefined()
+  expect(Role.USER).toBeDefined()
+  expect(Role.ADMIN).toBeDefined()
```
janpio: Is this enough testing to make sure Prisma Client actually works? Should it maybe try to do a query and confirm the data returned etc?

SevInf: This tests browser bundle, it does not actually contain client - just generated enums and version information.

bundlers/webpack-browser-custom-output/package.json
SevInf: Yes, I'll update it do dev version once prisma/prisma#13291 is merged

bundlers/webpack-browser-custom-output/index.test.js
janpio: Are the other tests in the `bundler` folder similar? Some of the CI pipeline certainly expects more, as also seen via the `schema.prisma` file here and so on and the tests failing because of missing engines etc.

bundlers/webpack-browser-custom-output/package.json
janpio: After these PRs pass on the integration version, I often run them on `dev` already just to see them to _actually_ fail on CI as well before the fix was merged :D

bundlers/webpack-browser-custom-output/index.test.js
SevInf: No, other bundler target node.js, where client is available. The whole purpose of this test is to test that bundling for browser also works (prisma/prisma#12296). Do you think it'll make sense create a separate category for those kinds of tests (`bundler-browser`?) to make that distinction clearer?

janpio: @millsp will know more if we already have something like this in another test. Generally the whole repo was set up to test if Prisma Client works in certain situations (databases, deployments and so on).

SevInf: @millsp was the one who sent me to this repo.

millsp: The best way to check if it's the browser build running is to do `new PrismaClient` and check that it throws an error saying `PrismaClient is unable to be run in the browser`.

SevInf: @janpio @millsp Updated the test to test for that exception

PR #2810, opened 2022-05-24
feat: Show nicer diff for `fetch-retry-and-confirm-version`

SevInf:
Uses `jest-diff` package which outputs the diff between two objects in
the same way jest-cli does.

Example output:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "@types/node": "14.18.18",
+    "jest-diff": "28.1.0",
     "ts-node": "10.7.0",
     "typescript": "4.6.4"
   },
--- a/package.json
+++ b/package.json
@@ -3,12 +3,12 @@
     "arg": "5.0.1",
     "fetch-retry": "5.0.2",
     "glob": "8.0.1",
+    "jest-diff": "28.1.0",
     "node-fetch": "2.6.7",
     "yaml": "2.0.0-11"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "@types/node": "14.18.18",
+    "jest-diff": "28.1.0",
```
janpio: Sure this is fine to just put as devDependency? Not super sure why the others (E.g. `node-fetch` which is also used in `utils/fetch-retry-and-confirm-version.ts`) are full dependencies...

utils/fetch-retry-and-confirm-version.ts
```diff
@@ -51,10 +48,7 @@ async function fetchRetry(args: FetchRetryArgs) {
   const data = await r.text()
 
   if (JSON.stringify(data) !== JSON.stringify(expectedData)) {
-    console.log('expected:')
-    console.log(JSON.stringify(expectedData))
-    console.log('but got:')
-    console.log(JSON.stringify(data))
+    console.log(diff(expectedData, data))
```
janpio: This is amazing, but let's still output the full strings as they are returned anyway to make it easier to copy/paste them to e.g. replace the "snapshot". (Maybe the diff first, and then the two explicit ones?)

janpio (approved):
Well, what a neat idea to use actually helpful tools rather than to make us copy/paste this locally for a diff :p 

Left optional suggestions/questions.

PR #2845, opened 2022-06-03
chore: updates [DPGA] 

millsp:


janpio (approved):
The non package.json (and lockfile) changes all look good and clean. Now you just need to get it to actually run the tests :D 

(Oh, this should not be an approval but just a normal review comment)

platforms-serverless/netlify-cli/test.sh
```diff
@@ -3,9 +3,9 @@
 set -eux
 
 if [ "$PRISMA_CLIENT_ENGINE_TYPE" == "binary" ]; then
-  BINARY_STRING=',"files":["edge","index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","runtime","schema.prisma"]'
+  BINARY_STRING=',"files":["index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","runtime","schema.prisma"]'
 else
-  BINARY_STRING=',"files":["edge","index-browser.js","index.d.ts","index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","runtime","schema.prisma"]'
+  BINARY_STRING=',"files":["index-browser.js","index.d.ts","index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","runtime","schema.prisma"]'
```
janpio: huh? why is this not present here?

millsp: Because we don't generate it in custom locations, unless you have `--data-proxy`. It is logically sound with our current implementation: we don't ever copy "throw" placeholders in custom locations, but they are there for the default imports always.

millsp: The experience is debatable of course, but I would need more time to change the logic to have placeholders just for edge client in custom locations. I'll see what I can do, if time permits.

janpio: Oh no, absolutely no need for that - but I was not aware this test uses a custom location. I think that makes the test for the files a bit useless here, or even wrong.

janpio: On the other hand, the "placeholder" should not have binaries, so I am not sure what is going on.

I am fine with you just making it pass with this change here. But maybe make a note that we need to look into this test here to really figure out what is going on.

millsp: These here aren't placeholders, they are the actual generated client in a custom location. That's why I said that it does not make sense right now in the code to have placeholders for `edge` in an actual generated client in custom locations. But we can change that after GA, of course.

dataproxy/vercel-edge-functions/middleware.ts
```diff
@@ -0,0 +1,17 @@
+import type { NextRequest } from 'next/server'
+import { NextResponse } from 'next/server'
+import { PrismaClient } from '@prisma/client/edge'
+
+export async function middleware(request: NextRequest) {
+  const prisma = new PrismaClient({})
+
+  // we use a transaction as it has caused issues with tracing
+  // `traceparent: undefined` was being passed to the engine
+  // but we ultimately want to test that queries are sent well
```
janpio: ?

millsp: It's an old comment from copy pasted code, it's not relevant anymore.

janpio:
- Seems Codesandbox is broken with deployment error, might be unrelated
- GCP Functions fails `"message": "Maximum number of functions in region us-central1 in project *** has been reached",`, so def unrelated
- Same for Firebase Functions: `⚠ functions: got "Quota Exceeded" error while trying to create projects/***/locations/us-central1/functions/e2e_firebase_test_2022_06_05_233227. Waiting to retry...`
- MacOS Azure SQL Server is most probably also unrelated, looks like the server is gone:
```
    Invalid `prisma.user.findUnique()` invocation in
    /Users/runner/work/ecosystem-tests/ecosystem-tests/databases-macos/sqlserver-azure-sql/index.test.js:11:36
       8 })
       9 
      10 it('should query the database', async () => {
    → 11   const data = await prisma.user.findUnique(
      Server has closed the connection.
```

.github/scripts/check-engines-cli.sh
```diff
@@ -84,7 +84,7 @@ elif [ $CLI_QUERY_ENGINE_TYPE == "library" ]; then
       qe_location2="node_modules\prisma\node_modules\engines\query_engine-windows.dll.node"
       ;;
   esac
-elif [ $CLI_QUERY_ENGINE_TYPE == "dataproxy" ]; then
+elif [ $CLI_QUERY_ENGINE_TYPE == "<dataproxy>" ]; then
```
janpio, line 87: What is this about? Just disabling the code without removing it?

janpio: Or reusing the old code to easily trigger that logic via the old env var instead of completely re-factoring ecosystem-tests?

janpio (approved):
Passing tests look great.

(Will now additionally have a manual look at the data proxy deployments)

Update: Generally looks good besides the things I mentioned via Slack.

.github/scripts/check-engines-cli.sh
millsp: > Or reusing the old code to easily trigger that logic via the old env var instead of completely re-factoring ecosystem-tests?

Yes, we can improve that later.

millsp:
Similar failures here https://github.com/prisma/ecosystem-tests/pull/2848

danstarns (approved):

community-generators/prisma-json-schema-generator/package.json
```diff
@@ -4,11 +4,11 @@
   "main": "index.js",
   "license": "MIT",
   "devDependencies": {
-    "prisma": "3.15.0-dev.51",
-    "prisma-json-schema-generator": "2.0.14"
+    "prisma": "3.15.0-integration-chore-data-proxy-ga-5.29",
+    "prisma-json-schema-generator": "2.0.13"
```
aqrln: Is this intentional?

aqrln (approved):

community-generators/prisma-json-schema-generator/package.json
millsp: Yes, I will remove it before merging.

PR #2853, opened 2022-06-07
chore: debug for codesandbox + remove previewFeatures flags

Jolg42:
Increased timeout and debug for codesandbox and cleanup of previewFeatures

### DIFF
--- a/platforms/codesandbox/deploy.ts
+++ b/platforms/codesandbox/deploy.ts
@@ -36,16 +36,19 @@ async function fetchWithPuppeteer(endpoint) {
   const page = await browser.newPage()
   await page.setDefaultNavigationTimeout(0)
   await page.goto(endpoint)
-  await page.waitForTimeout(10000)
+  await page.waitForTimeout(20_000)
   const screenshot = await page.screenshot()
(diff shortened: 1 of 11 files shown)


platforms/codesandbox/deploy.ts
```diff
@@ -36,16 +36,19 @@ async function fetchWithPuppeteer(endpoint) {
...
+  await page.waitForTimeout(20_000)
   const screenshot = await page.screenshot()
   fs.writeFileSync('image.png', screenshot as Buffer)
   await browser.close()
 
   const r = await fetch(endpoint)
+  const body = await r.text()
   try {
-    await r.json()
+    const bodyAsJSON = JSON.parse(body)
```
janpio, line 47: When you did `r.text()` you can not use `r.json()` any more to get it already parsed?

janpio (approved):

platforms/codesandbox/deploy.ts
Jolg42: Indeed!

Jolg42: Or you will get an error

PR #2872, opened 2022-06-15
chore: test data proxy on regular vercel functions

millsp:


.github/workflows/test.yaml
```diff
@@ -800,6 +800,7 @@ jobs:
           - nodejs
           - cloudflare-workers
           - vercel-edge-functions
+          - vercel-cli
```
janpio: Let's rename this to:
```suggestion
          - vercel-serverless-functions-via-cli
```

janpio: (We might want to add `via git` later or other deployment mechanisms)

dataproxy/vercel-cli/api/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+const prisma = new PrismaClient({})
+
+export default async (req, res) => {
+  const data = await prisma.$transaction([
+    prisma.user.findFirst(),
+    prisma.user.findMany()
+  ])
+
+  const json = JSON.stringify({ data })
```
janpio: Why simplified script with no creation and file checking here?

dataproxy/vercel-cli/finally.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -eu
+
+yarn vercel logs https://e2e-dataproxy-vercel-cli.vercel.app --token=$VERCEL_TOKEN --scope=$VERCEL_ORG_ID
```
janpio: Requesting this URL directly currently download the JS source of some file:
```
/// <reference types="@types/jest" />

import fetch from 'node-fetch'

describe('use data proxy', () => {
  test('fetch response', async () => {
    console.debug(process.env.DEPLOYMENT_URL!)
    
    const response = await fetch(process.env.DEPLOYMENT_URL!)
    
    const bodyAsText = await response.text()
    console.debug(bodyAsText)
    
    const jsonData = JSON.parse(bodyAsText)
    expect(jsonData).toMatchInlineSnapshot(`
Object {
  "data": Array [
    Object {
      "email": "Eleanore_Feest@yahoo.com",
      "name": "Autumn",
      "user_id": 1,
    },
    Array [
      Object {
        "email": "Eleanore_Feest@yahoo.com",
        "name": "Autumn",
        "user_id": 1,
      },
      Object {
        "email": "Ezra.Borer@gmail.com",
        "name": "Marguerite",
        "user_id": 2,
      },
      Object {
        "email": "Dustin.Erdman31@gmail.com",
        "name": "Jana",
        "user_id": 3,
      },
      Object {
        "email": "Merle3@gmail.com",
        "name": "Pink",
        "user_id": 4,
      },
      Object {
        "email": "Fredrick.Howe@yahoo.com",
        "name": "Oren",
        "user_id": 5,
      },
      Object {
        "email": "Conrad_Lowe@hotmail.com",
        "name": "Sigmund",
        "user_id": 6,
      },
      Object {
        "email": "Nicola_Hessel10@hotmail.com",
        "name": "Toy",
        "user_id": 7,
      },
      Object {
        "email": "Fredrick_Conroy76@hotmail.com",
        "name": "Brooklyn",
        "user_id": 8,
      },
      Object {
        "email": "Norris40@hotmail.com",
        "name": "Cassandre",
        "user_id": 9,
      },
      Object {
        "email": "Osborne96@gmail.com",
        "name": "Jazmyne",
        "user_id": 10,
      },
    ],
  ],
}
`)
  }, 30000)
})

```

dataproxy/vercel-cli/prisma/schema.prisma
```diff
@@ -0,0 +1,31 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATAPROXY_VERCEL_CLI_DB_URL")
```
janpio: Where is this database hosted?

dataproxy/vercel-cli/run.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+set -eu
+
+yarn
+
+yarn -s vercel --token=$VERCEL_TOKEN --build-env PRISMA_GENERATE_DATAPROXY="true" --env DATAPROXY_VERCEL_CLI_DB_URL="$DATAPROXY_VERCEL_CLI_DB_URL" --prod --scope=$VERCEL_ORG_ID --confirm --force 1> deployment-url.txt
```
janpio: We are only using the env var in this test setup, not `--data-proxy` I guess?

Edit: Oh yeah, obviously - we added this env var because the other thing did not work :D

dataproxy/vercel-cli/finally.sh
millsp: Yeah, the jest test got uploaded, I'll see if adding to gitignore helps

dataproxy/vercel-cli/prisma/schema.prisma
millsp: It's the same db as all the other deployments, so provisioned by us and then put behind the data proxy

dataproxy/vercel-cli/run.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+fi
+
+# Check the Vercel Build Logs for "Generated Prisma Client"
+if echo "${OUTPUT}" | grep -q 'Generated Prisma Client'; then
+  echo 'Prisma Client Was Successfully Generated'
+else
+  echo "Prisma Client Was NOT GENERATED"
+  exit 1
+fi
```
janpio: We should also check for `| dataproxy` here maybe to make sure a Data Proxy one was generated along the way somewhere.

dataproxy/vercel-cli/api/index.js
millsp: We don't yet have data creation tests, we would need to provision a db and hook the data proxy to it, otherwise tests that run at the same time could write to the same db (eg. integration, patch-dev, dev), I think?

dataproxy/vercel-cli/finally.sh
janpio: Weird that it chooses that as the thing to return though as well.

dataproxy/vercel-cli/prisma/schema.prisma
janpio: Ah ok, so this is a data proxy specific one?

That also explains why we only read and not write I guess (other comment).

dataproxy/vercel-cli/api/index.js
janpio: Yes - I did not make the connection that this is all the same database for the dataproxy tests.
We should have an issue to expand that. Writing is kind of important :D

millsp: https://github.com/prisma/ecosystem-tests/issues/2874

PR #2968, opened 2022-07-25
test: add tracing

danstarns:
Tests that tracing works within another traced application.

Spins up a fastify web server, performs a request against that server, and then asserts that those traces and Prisma traces are nested correctly within each other.


core-features/tracing/run.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -eux
+
+yarn install
+yarn prisma generate
+
+docker-compose up -d
```
Jolg42: Could you move this to a new `prepare.sh` file? It will be executed automatically by the setup first.

core-features/tracing/index.test.js
```diff
@@ -0,0 +1,447 @@
...
+            {
+              \\"key\\": \\"otel.library.name\\",
+              \\"type\\": \\"string\\",
+              \\"value\\": \\"@opentelemetry/instrumentation-fastify\\"
+            },
+            {
+              \\"key\\": \\"otel.library.version\\",
+              \\"type\\": \\"string\\",
+              \\"value\\": \\"0.25.0\\"
+            },
```
Jolg42: Is `1.4.0` for:
    `"@opentelemetry/exporter-jaeger": "^1.4.0",`

Is `0.25.0` for:
    `"@opentelemetry/instrumentation-fastify": "^0.25.0",`

These versions in the snapshot will be annoying later to maintain as any change will break the test.
Is there another way to test this? or remove the versions / normalize them?

core-features/tracing/run.sh
```diff
@@ -0,0 +1,10 @@
...
+
+node app.js
```
Jolg42: Tests are currently failing with 
```
+ node app.js
Warning: Attempt 2 failed. Reason: Timeout of 600000ms hit
```

```suggestion
node app.js &
```
Adding `&` should not wait indefinitely = background process (see example like https://linuxhint.com/run-background-command-bash/)

core-features/tracing/index.test.js
danstarns: Good idea! Removed.

core-features/tracing/index.test.js
```diff
@@ -0,0 +1,390 @@
...
+          'duration',
+          'warnings',
+          'processID',
+          'logs',
+        ]
+        if (ignoredKeys.includes(key)) {
+          return undefined
+        }
+
+        if (key === 'value' && ['telemetry.sdk.version', 'otel.library.version'].includes(this.key)) {
```
Jolg42, line 70:
```suggestion
	// Remove value for the following keys
	// they contain a version that can change
	// removing them makes the test easier to maintain
        if (key === 'value' && ['telemetry.sdk.version', 'otel.library.version'].includes(this.key)) {
```

core-features/tracing/prisma/schema.prisma
```diff
@@ -0,0 +1,15 @@
+generator client {
+    provider        = "prisma-client-js"
+    previewFeatures = ["tracing"]
+}
+
+datasource db {
+    provider = "postgresql"
+    url      = "DATABASE_URL"
```
Jolg42:
```suggestion
    url      = env("DATABASE_URL")
```
That should do it

core-features/tracing/app.js
```diff
@@ -0,0 +1,25 @@
+require('./tracing')
```
janpio, line 2:
```suggestion
// enable Otel tracing in this app
require('./tracing')
```

core-features/tracing/tracing.js
```diff
@@ -0,0 +1,25 @@
...
+const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node')
+const { SimpleSpanProcessor } = require('@opentelemetry/sdk-trace-base')
+const { PrismaInstrumentation } = require('@prisma/instrumentation')
+const { FastifyInstrumentation } = require('@opentelemetry/instrumentation-fastify')
+const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http')
+const { JaegerExporter } = require('@opentelemetry/exporter-jaeger')
+
+const provider = new NodeTracerProvider({
+  resource: new Resource({
+    [SemanticResourceAttributes.SERVICE_NAME]: 'prisma-tracing',
```
janpio:
```suggestion
    [SemanticResourceAttributes.SERVICE_NAME]: 'ecosystem-tests-core-features-tracing',
```

millsp:
Ping @danstarns, we might need to do some updates here

Jolg42:
ping @danstarns @millsp, found this PR which looks almost done, would be good to finish it

core-features/tracing/index.test.js
```diff
@@ -0,0 +1,393 @@
...
+
+    const jeagerResponse = await axios({
+      method: 'GET',
+      url: 'http://localhost:16686/api/traces?service=ecosystem-tests-core-features-tracing',
+    })
+    expect(jeagerResponse.status).toEqual(200)
+
+    const [trace] = jeagerResponse.data.data
+    const [rootSpan, ...restOfSpans] = trace.spans
+    const traceTree = buildTree({ span: rootSpan }, restOfSpans)
```
Jolg42, line 47: @danstarns I think you might need to order the JSON first to have a stable snapshot below? or find another approach like snapshotting each value separately

janpio (changes requested):
The snapshot is from a SQLite run, but the test runs PostgreSQL.

core-features/tracing/index.test.js
janpio: The functional tests now implement a more advanced `buildTree`, although that looks a bit dangerous that it might actually delete useful information: https://github.com/prisma/prisma/blob/f46bca1625018ebf94ee1b3242cc40c3539c49ea/packages/client/tests/functional/tracing/tests.ts#L25-L74

mcbethio:
@millsp this has a milestone of 4.8.0 but not started yet - I assume the milestone should be removed?

millsp:
@mcbethio Work was started but we never got the time to finish it.

janpio:
We don't usually update milestones when nothing happens (or remove them when work stops). We have loads of issues and PRs with older milestones.

Issue #1921, opened 2021-07-10
Use yarn resolutions for redwood dependency override


millsp:
Internal https://prisma-company.slack.com/archives/CL0AJEVSA/p1625882635040500

janpio:
Should already be possible in the update script as we were using that before.

If we can avoid it, we should though as it doubles the complexity and was a source of endless problems in the past.

Jolg42:
Note: we currently use the override in 
https://github.com/prisma/ecosystem-tests/blob/dev/platforms-serverless-vercel/vercel-with-redwood/api/package.json#L9-L12
https://github.com/prisma/ecosystem-tests/blob/dev/platforms-serverless-vercel/vercel-with-redwood/package.json#L16-L19

And we have this script (https://github.com/prisma/ecosystem-tests/blob/dev/platforms-serverless-vercel/vercel-with-redwood/patch-package-json.js
) that seems to update the version with the version from
https://github.com/prisma/ecosystem-tests/blob/dev/.github/prisma-version.txt

Jolg42:
We now have an issue about check-for-update doesn't bump resolutions in package.json -> https://github.com/prisma/ecosystem-tests/issues/3041

PR #3004, opened 2022-08-10
feat: isolate library & binary engine testing

Jolg42:
if we can isolate tests we can then, have Slack notifications, and not me looking after it regularly 😄


janpio:
You are owner of the team at https://vercel.com/teams/prisma-orm/ - so should be able to fix.
Make sure to use current credentials, we moved the deployment to a new account some time ago.

Jolg42:
@janpio These errors happen after a fresh login which is concerning...

I'll clear and try again later, if we can isolate tests we can then have Slack notifications and not me looking after it regularly 😄

bundlers/parcel/src/index.js
```diff
@@ -3,8 +3,6 @@ import { PrismaClient } from '@prisma/client'
 const client = new PrismaClient()
 
 export async function main() {
-  await client.user.deleteMany({})
```
janpio: Why remove?

bundlers/rollup/src/index.js
```diff
@@ -3,8 +3,6 @@ import { PrismaClient } from '@prisma/client'
 const client = new PrismaClient()
 
 export async function main() {
-  await client.user.deleteMany({})
```
janpio: Why remove?

bundlers/webpack/src/index.js
```diff
@@ -3,8 +3,6 @@ import { PrismaClient } from '@prisma/client'
 const client = new PrismaClient()
 
 export async function main() {
-  await client.user.deleteMany({})
```
janpio: Why remove?

databases/heroku-pgbouncer-buildpack/index.js
```diff
@@ -14,8 +14,6 @@ const app = express()
 const port = process.env.PORT || 3000
 
 app.get('/', async (req, res) => {
-  await clientWithQueryStringParam.user.deleteMany({})
```
janpio: Why remove?

databases/heroku-pgbouncer-buildpack/index.js
```diff
@@ -34,18 +32,20 @@ app.get('/', async (req, res) => {
   })
   const users = await clientWithQueryStringParam.user.findUnique({
     where: {
-      id,
+      id: createUser.id,
     },
   })
+  const deleteUser = await clientWithQueryStringParam.user.delete({
+    where: { id: 1 },
```
janpio: Why delete different user than created?

platforms-serverless-vercel/vercel-cli/api/index.js
```diff
@@ -2,21 +2,16 @@ const { PrismaClient, Prisma } = require('@prisma/client')
 const client = new PrismaClient()
 
 export default async (req, res) => {
-  await client.user.deleteMany({})
```
janpio: Why remove?

databases/heroku-pgbouncer-buildpack/index.js
Jolg42: Typo, nice catch
```suggestion
    where: { id: createUser.id },
```

databases/heroku-pgbouncer-buildpack/index.js
Jolg42: Because it's deleting all users, and we can make this isolated across the runs and only create / delete one user.

Same answer to the other comments

platforms-serverless-vercel/vercel-with-redwood/api/upgrade.sh
```diff
@@ -2,4 +2,4 @@
 
 set -eu
 
-yarn policies set-version 1.18.0
+yarn policies set-version 1.22.19
```
Jolg42, line 5: Remove this warning and it's the version used by most people (1.22)
```
prisma:warn Your yarn has version 1.18.0, which is outdated. Please update it to 1.19.2 or newer in order to use Prisma.
```

databases/heroku-pgbouncer-buildpack/index.js
janpio: Hm ok, I think this pattern is used in many other places  but might make sense to universally change.

platforms-serverless-vercel/vercel-with-redwood/api/package.json
```diff
@@ -7,7 +7,7 @@
     "@redwoodjs/graphql-server": "2.2.0"
   },
   "resolutions": {
-    "prisma": "3.8.0-dev.23",
-    "@prisma/client": "3.8.0-dev.23"
+    "prisma": "4.3.0-dev.8",
+    "@prisma/client": "4.3.0-dev.8"
```
Jolg42, line 11: That was very outdated

Jolg42: Seems we have nothing that bumps it automatically

janpio: That's quite a problem then, seems this project is not working great.

databases/heroku-pgbouncer-buildpack/index.js
Jolg42: Did a revert, better to do in a separate PR

janpio:
Rnadom thought: Please double check that all the jobs that you split are using the "automatic" database "provisioning" step in CI, and do not have a custom database set in the deployment platform. Otherwise this might be a bit more complicated than it looks on first sight.

Jolg42:
@janpio Let's merge this, we can do others incrementally

janpio:
Did you look into the databases thing?

Jolg42:
@janpio the vercel deployments have a database env var set on their side. So there is nothing else to do here I think.

I think to avoid the constraints issues, we can drop `@unique` on email and remove the fixed id used in the tests. (= still use 1 database for both)

platforms-serverless-vercel/vercel-with-redwood/api/package.json
janpio: Best create a separate issue (if it does not exist yet)

platforms-serverless-vercel/vercel-with-redwood/api/upgrade.sh
janpio: What is this used/needed for? Wort adding a comment? Reviewing this I have 0 context.

platforms-serverless-vercel/vercel-with-redwood/package.json
```diff
@@ -14,8 +14,8 @@
     "vercel": "27.4.0"
   },
   "resolutions": {
-    "prisma": "3.8.0-dev.23",
-    "@prisma/client": "3.8.0-dev.23"
+    "prisma": "4.3.0-dev.8",
+    "@prisma/client": "4.3.0-dev.8"
```
janpio, line 18: Probably also worth an issue if this is relevant.

platforms-serverless-vercel/vercel-with-redwood/run.sh
```diff
@@ -2,25 +2,31 @@
 
 set -eu
 
-yarn policies set-version 1.18.0
+yarn policies set-version 1.22.19
```
janpio, line 5: Also comment

platforms-serverless-vercel/vercel-with-redwood/upgrade.sh
```diff
@@ -2,4 +2,4 @@
 
 set -eu
 
-yarn policies set-version 1.18.0
+yarn policies set-version 1.22.19
```
janpio, line 5: comment

platforms-serverless-vercel/vercel-with-redwood/web/upgrade.sh
```diff
@@ -2,4 +2,4 @@
 
 set -eu
 
-yarn policies set-version 1.18.0
+yarn policies set-version 1.22.19
```
janpio, line 5: comment

janpio:
> @janpio the vercel deployments have a database env var set on their side. So there is nothing else to do here I think.

So I assume the new deployment just got the same database set as the old one now?
It would be better if we could "inject" that env var from the script and make them use the same, auto generated database as most other tests use. (But not necessary for merge here, but please issue)

janpio (approved):

platforms-serverless-vercel/vercel-with-redwood/api/upgrade.sh
Jolg42: Maybe we don't even need this anymore, I'll try removing them in a PR

janpio: I honestly do not know what it was for.

Jolg42: We probably had an issue with some version of yarn at some point and we had this use this specific version workaround

Jolg42: Did a revert, turned out to break in check-for-update, added a comment
https://github.com/prisma/ecosystem-tests/commit/88e5bca3b7cd63dd7cd5e623ac3a3953274328e2

platforms-serverless-vercel/vercel-with-redwood/api/package.json
Jolg42: I wanted to create an issue, but it seems to work mostly as expected, see https://github.com/prisma/ecosystem-tests/issues/1921#issuecomment-1223814773

The patch script works (tried it), and updates the resolutions before testing.
But we do not commit this as we only commit during `check-for-update`

PR #3005, opened 2022-08-10
chore: use cache for yarn

Jolg42:
And random cleanup

### DIFF
--- a/.github/workflows/check-for-update.yaml
+++ b/.github/workflows/check-for-update.yaml
@@ -17,8 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: use node 14
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
 
@@ -34,8 +33,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: use node 14
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
 
@@ -51,8 +49,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: use node 14
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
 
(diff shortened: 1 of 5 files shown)


.github/workflows/check-for-update.yaml
```diff
@@ -17,8 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: use node 14
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
```
janpio, line 20: version being used still explicit in the logs?

.github/workflows/invoke-platforms-lambda-function.yml
```diff
@@ -1,23 +1,26 @@
...
-      
+
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
-      - run: yarn
+        with:
+          cache: 'yarn'
+      - name: Install Dependencies
+        run: yarn install
```
janpio, line 25: difference between just `yarn` and `yarn install`?

Jolg42: Just having one explicit everywhere, it's the same but easier to search

.github/workflows/optional-test.yaml
```diff
@@ -33,6 +33,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
+        with:
+          cache: 'yarn'
```
janpio, line 37: Is this lacking an install deps step?

janpio (commented):
Can we somehow extract these super common and repeated steps now maybe? I have some vagues recollection of GH Actions adding some composability recently for theri yaml.

.github/workflows/check-for-update.yaml
Jolg42: Only if you open the section. But not needed and can drift.
For tests where we explicitly want to test the node version we use a matrix and it's visible

.github/workflows/optional-test.yaml
Jolg42: Apparently no yarn install is needed here because it has no dependencies, so cache is not needed as well

Jolg42:
We can check, it could definitely help but not sure it's what we want here (the feature might not do what we want)

PR #3070, opened 2022-09-01
feat: move mongodb-azure-cosmosdb to optional tests

janpio:
This moves the currently failing test `databases/mongodb-azure-cosmosdb` from normal tests to optional tests, so that failure does not actually fail the build and just shows up as red x in the list if you look at it.


Jolg42 (commented):
Thanks!

Jolg42 (approved):

PR #3342, opened 2023-01-03
ci: add debian-latest-arm64-openssl-1.1.x to docker workflow

jkomyno:
Closes https://github.com/prisma/prisma-private/issues/205.

As it turns out, the `debian-latest-arm64-openssl-1.1.x` Docker test was failing for `prisma@4.8.0`, but is succeeding for the latest `dev`-version of `prisma`. 🎉

### DIFF
--- a/docker/debian-latest-arm64-openssl-1.1.x/run.sh
+++ b/docker/debian-latest-arm64-openssl-1.1.x/run.sh
@@ -5,7 +5,7 @@ export DEBUG="*"
 
 yarn install
 
-DOCKER_PLATFORM_ARCH="linux/amd64"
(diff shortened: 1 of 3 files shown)


docker/debian-latest-arm64-openssl-1.1.x/run.sh
```diff
@@ -5,7 +5,7 @@ export DEBUG="*"
 
 yarn install
 
-DOCKER_PLATFORM_ARCH="linux/amd64"
```
jkomyno, line 8: The tests passes regardless of this change as we're using a base Docker image which is implicitly tagged for `linux/arm64`. However, let's keep things consistent.

Jolg42 (approved):

PR #3343, opened 2023-01-04
feat: docker os support followup

### LINKED ISSUE #14017: Create CI worker without a proper `openssl` or `libc` installation to improve snapshot testing suite
Currently, we're unable to easily reproduce errors that only emerge in environments without version of `openssl` or `libc` that we support. The standard approach we take is to wait for users to report issues (which may simply be due to unclear error messages), spin up a VM droplet with flaky system libraries installations with @pantharshit00's help, reproduce the problem and try to fix it from there.

As we don't have such an environment accessible via CI, it's currently impossible to perform snapshot testing on the error messages we trigger in those situations. For instance, this resulted in https://github.com/prisma/prisma/pull/13736 getting merged without anyone noticing a missing space (`" "`) character after `"Details:"`.

Screenshot 2022-06-27 at 12 55 40

## Proposed Solution

Here are two alternatives:

1. Create a dedicated CI worker that mirrors an improper `openssl` or `libc` installation (can this be done easily in Github Actions?)
2. Use a standard CI worker that opens an `ssh` connection to an existing VM droplet configured as above, and runs the test there (more convoluted and tricky to get right)

jkomyno:
This PR:
- adds new `docker` tests:
  - [x] `debian-stretch-amd64-openssl-1.1.x`
  - [ ] `distroless-bullseye-amd64-openssl-1.1.x` (currently failing although I have used it to solve https://github.com/prisma/prisma/issues/16205, low priority, we can fix it in a later PR)
  - [x] `redhat-ubi9-amd64-openssl-3.0.x`
- adds new `docker-unsupported` tests, which are stored in `docker/_fail/*` and are expected to fail
  - [x] `alpine-3.16-arm64-openssl-1.1.x`
  - [x] `alpine-3.16-arm64-openssl-1.1.x-with-libc`
  - [x] `alpine-3.17-arm64-openssl-3.0.x`
  - [x] `alpine-3.17-arm64-openssl-3.0.x-with-libc`
  - [x] `debian-buster-amd64-openssl-1.1.x` (no `openssl` installed)
  - [ ] `distroless-bullseye-amd64-openssl-1.1.x` (no `libz` installed)
- use yarn on the host system (i.e., not Docker) only after the Docker test has run
- polishes minor details / comments

It also closes https://github.com/prisma/prisma/issues/14017.

## Review Filters

- [Relevant scripts only](https://github.com/prisma/ecosystem-tests/pull/3343/files?file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&show-deleted-files=true&show-viewed-files=true): File filter > Filter by extension > Check `.yaml`, `.sh`
- [Dockerfiles only](https://github.com/prisma/ecosystem-tests/pull/3343/files?file-filters%5B%5D=No+extension&show-deleted-files=true&show-viewed-files=true): File filter > Filter by extension > Check `No extension`
- [Markdowns only](https://github.com/prisma/ecosystem-tests/pull/3343/files?file-filters%5B%5D=.md&show-deleted-files=true&show-viewed-files=true): File filter > Filter by extension > Check `.md`
- [All relevant changes](https://github.com/prisma/ecosystem-tests/pull/3343/files?file-filters%5B%5D=.md&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension&show-deleted-files=true&show-viewed-files=true): File filter > Filer by extension > Check `.md`, `.yaml`, `.sh`, `No extension`


.github/scripts/check-folders.js
```diff
@@ -45,7 +44,7 @@ async function main() {
   // Get all relevant folders that _should_ appear in workflows
   const folders = glob
     .sync('**/package.json', {
-      ignore: ['**/node_modules/**', '**/custom-engines/**'],
+      ignore: ['**/node_modules/**', '**/custom-engines/**', 'docker/_fail/**'],
```
jkomyno: **Note**: `docker/_fail` tests are ignored by `check-folders.js`, which is used when linting ecosystem-tests.

.github/workflows/test.yaml
```diff
@@ -242,6 +245,51 @@ jobs:
...
+      - uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: "arm64"
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
```
jkomyno, line 276: Roughly same steps as the `docker` workflow, but `node` is not needed (as no Slack notification is sent on failure).

.github/workflows/test.yaml
```diff
@@ -242,6 +245,51 @@ jobs:
         if: failure()
         run: bash .github/slack/notify-failure.sh ${{ github.job }} ${{ matrix.feature }}
 
+  docker-failing:
```
```suggestion
  # These tests are expected to fail
  # Meaning a successful run actually shows a non zero exit code in the logs
  docker-failing:
```
Random comment

docker/README.md
```diff
@@ -5,16 +5,17 @@ This folder builds and run Prisma on a series of different system configurations
...
 ```
   
 where:
-- `MAYBE_FAIL` is either `fail-` (when running the image would result in an error) or `` otherwise
 - `LINUX_DISTRO` is the name of the Linux distribution (e.g. `ubuntu`, `debian`, `alpine`)
 - `VERSION` is the version of the Linux distribution (e.g. `22.04`, `bullseye`, `latest`)
 - `ARCHITECTURE` is the architecture of the Linux distribution (e.g. `amd64`, `arm64`)
 - `OPENSSL_VERSION` is the main version of OpenSSL used by the system (e.g. `1.1.x`, `3.0.x`)
 
+Tests that are expected to fail are stored in the [`fail-`](./_fail/) folder.
```
Jolg42:
```suggestion
Tests that are expected to fail are stored in the [`_fail`](./_fail/) folder.
```

docker/alpine-latest-amd64-openssl-3.0.x/run.sh
```diff
@@ -30,3 +28,5 @@ docker run -p 3000:3000 \
   "${PRISMA_DOCKER_IMAGE_NAME}" &
 
 sleep 15
+
+yarn install
```
Jolg42, line 32: Any specific reason why this moved?

Jolg42 (approved):
Looks good to me 👍🏼 

See comments before merging

Also I'm unsure if we should run the failing docker tests for each new prisma version (we could change that later)

docker/alpine-latest-amd64-openssl-3.0.x/run.sh
jkomyno: `yarn install` is run on the host machine (so, not Docker) and will always install the engines of Prisma that match the Github Actions' runner. If `yarn install` is run before `docker buildx ...`, Docker takes longer to build and run its image because the local context is heavier (due to `node_modules`). This local installation of Prisma is only needed after `./run.sh` has run anyway, so we're not changing any behavior, we're just avoiding slowing down the pipeline (which is also particularly handy locally)

Jolg42: Makes a lot of sense! thanks!

.github/scripts/check-folders.js
```diff
@@ -23,7 +23,10 @@ async function main() {
     'packagers/yarn3-workspaces-pnp/packages/sub-project-2', // We don't want to include the workspace folders in the matrix
     'platforms/aws-graviton/code', // aws-graviton doesn't have package.json at root but is included
     'platforms/m1-macstadium/code', // m1-macstadium doesn't have package.json at root but is included
-    'docker/_fail-debian-buster-amd64-openssl-1.1.x', // docker/_fail-* is test that fails once run
+    
+    // commenting to let https://github.com/prisma/ecosystem-tests/pull/3343 succeed, lowish-priority to fix
+    'docker/_fail/distroless-bullseye-amd64-openssl-1.1.x',
+    'docker/distroless-bullseye-amd64-openssl-1.1.x',
```
janpio: not actually commented out

jkomyno: Right, it should be replaced with "// ignoring to let ..."

.github/scripts/test-project.sh
```diff
@@ -92,10 +116,8 @@ if [ $code -eq 0 ]; then
   echo "finished test.sh (code $code)"
   echo ""
   echo "-----------------------------"
-fi
 
-# confirm existence of correct engine
-if [ $code -eq 0 ]; then
```
janpio, line 98: Would this not also run this for all other failing test runs now?

.github/workflows/test.yaml
janpio: I suggest to comment out the unneeded things with a note but leave them, or sooner or later someone who cleans up the workflow file and wants everything to look the same will add it back.

.github/workflows/test.yaml
```diff
@@ -242,6 +245,53 @@ jobs:
         if: failure()
         run: bash .github/slack/notify-failure.sh ${{ github.job }} ${{ matrix.feature }}
 
+  # These tests are expected to fail
+  # Meaning a successfull run actually shows a non zero exit code in the logs
+  docker-failing:
+    needs: [detect_jobs_to_run]
+    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['docker'] == true }}
```
janpio: Hm, does that follow the usual pattern?
Wouldn't we usually have a top level `docker-failing` folder instead?

This is complicated enough, let's not introduce more special cases if possible.

docker/README.md
janpio: This would need to be updated when https://github.com/prisma/ecosystem-tests/pull/3343/files#r1062728565 is taken into account.

.github/scripts/test-project.sh
jkomyno: Prior to this PR, there were two consecutive , non-nested`if [ $code -eq 0 ]` checks.
The "confirm existence of correct engine" is now moved under the `elif [ $code -eq 0 ]; then` block in `test-project.sh`, after the `test.sh` part. No behavioral change for other test suites

docker/alpine-latest-amd64-openssl-3.0.x/run.sh
janpio: What do we need the local Prisma for at all?

jkomyno: For things unrelated to this PR. From what I have read skimming around, for reporting the version of Prisma installed and for running some js scripts after each test

janpio: Might not really make sense in Docker context. Probably there is a denylist to run those...

But indeed not necesary to fix right now.

.github/workflows/test.yaml
jkomyno: That's a good point, done. I see a couple of errors in the CI, will quickly solve after dinner

jkomyno: Fixed. Two optional tests are failing, which were failing before this PR (and my previous other PRs as well)

janpio: Yep, can confirm those are expected.
(We might want to split out "optional" and "expected failing" tests some time...)

.github/scripts/check-folders.js
```diff
@@ -45,7 +49,7 @@ async function main() {
   // Get all relevant folders that _should_ appear in workflows
   const folders = glob
     .sync('**/package.json', {
-      ignore: ['**/node_modules/**', '**/custom-engines/**'],
+      ignore: ['**/node_modules/**', '**/custom-engines/**', 'docker/_fail/**'],
```
janpio: Is this still needed?

docker-failing/README.md
```diff
@@ -0,0 +1,21 @@
+# Docker: OS Support (failing tests)
```
janpio: Late, but: 
Do the tests fail in CI? 
Or does the script return a failing status code and that is expected by the test and it succeeds?

`docker-unsupported` would the potentially be the better name.

janpio:
Hmm, just noticed this in the CI output of one of the `docker-failing` runs:
https://github.com/prisma/ecosystem-tests/actions/runs/3849578290/jobs/6558754456#step:7:70
```
+ npx prisma db push --accept-data-loss --skip-generate --schema=docker-failing/alpine-3.16-arm64-openssl-1.1.x/prisma/schema.prisma
npx prisma db push --accept-data-loss --skip-generate --schema=docker-failing/alpine-3.16-arm64-openssl-1.1.x/prisma/schema.prisma
npm WARN exec The following package was not found and will be installed: prisma@4.8.1
```
We should double check if that is normal of if the moving of the install command potentially impacted this.

janpio (approved):
Tests themselves look _great_ and show nicely how Prisma currently fails in certain environments 👍

janpio (changes requested):
Actually, this one is not happy and fails to early and for an unexpected reason: https://github.com/prisma/ecosystem-tests/actions/runs/3849578290/jobs/6558754809#step:7:97

docker-failing/README.md
jkomyno: .> Hmm, just noticed this in the CI output of one of the `docker-failing` runs: https://github.com/prisma/ecosystem-tests/actions/runs/3849578290/jobs/6558754456#step:7:70
> 
> ```
> + npx prisma db push --accept-data-loss --skip-generate --schema=docker-failing/alpine-3.16-arm64-openssl-1.1.x/prisma/schema.prisma
> npx prisma db push --accept-data-loss --skip-generate --schema=docker-failing/alpine-3.16-arm64-openssl-1.1.x/prisma/schema.prisma
> npm WARN exec The following package was not found and will be installed: prisma@4.8.1
> ```
> 
> We should double check if that is normal of if the moving of the install command potentially impacted this.

That's happening outside of Docker, and both the `package.json` and `yarn.lock` of `alpine-3.16-arm64-openssl-1.1.x` is mentioning `prisma@4.9.0-dev.30`. I wouldn't worry about it

jkomyno:
> Actually, this one is not happy and fails to early and for an unexpected reason: https://github.com/prisma/ecosystem-tests/actions/runs/3849578290/jobs/6558754809#step:7:97

Nice catch, I hadn't updated all the paths passed to `docker build`. It's fixed now

docker-failing/README.md
janpio: Does it in current other workflow runs (outside this PR) already use 4.8.1? Then this is just a general bug/problem.

.github/scripts/check-folders.js
```diff
@@ -23,7 +23,11 @@ async function main() {
     'packagers/yarn3-workspaces-pnp/packages/sub-project-2', // We don't want to include the workspace folders in the matrix
     'platforms/aws-graviton/code', // aws-graviton doesn't have package.json at root but is included
     'platforms/m1-macstadium/code', // m1-macstadium doesn't have package.json at root but is included
-    'docker/_fail-debian-buster-amd64-openssl-1.1.x', // docker/_fail-* is test that fails once run
+    
+    // ignoring docker distroless to let https://github.com/prisma/ecosystem-tests/pull/3343 succeed,
+    // lowish-priority to fix
```
janpio:
```suggestion
    // ignoring docker distroless as those are not actually run yet
```

docker-unsupported/README.md
```diff
@@ -0,0 +1,21 @@
+# Docker: OS Support (unsupported scenarios)
+
+This folder builds and run Prisma on a series of different unsupported system configurations and architectures, described by `Dockerfile`s.
```
janpio:
```suggestion
This folder builds and runs Prisma on a series of different unsupported system configurations and architectures, described by `Dockerfile`s.
```

docker-unsupported/alpine-3.16-arm64-openssl-1.1.x-with-libc/README.md
```diff
@@ -0,0 +1,16 @@
+# OS support: fail-alpine-3.16-arm64-openssl-1.1.x-with-libc
+
+- Base Docker image: `node:lts-alpine3.16`
+- OS: Linux Alpine 3.16
+- Arch: arm64 (aarch64)
+- OpenSSL location: `/lib/libssl.so.1.1`
+- Binary target: `linux-arm64-openssl-1.1.x`
+
+Prisma doesn't support the `arm64` architecture for Linux Alpine.
```
janpio: Do we have an issue that tracks that which we can link here? Then it will be easy to crossreference.

docker-unsupported/alpine-3.16-arm64-openssl-1.1.x/README.md
```diff
@@ -0,0 +1,16 @@
+# OS support: fail-alpine-3.16-arm64-openssl-1.1.x-with-libc
+
+- Base Docker image: `node:lts-alpine3.16`
+- OS: Linux Alpine 3.16
+- Arch: arm64 (aarch64)
+- OpenSSL location: `/lib/libssl.so.1.1`
+- Binary target: `linux-arm64-openssl-1.1.x`
+
+Prisma doesn't support the `arm64` architecture for Linux Alpine.
```
janpio: Link?

docker-unsupported/alpine-3.16-arm64-openssl-1.1.x/README.md
```diff
@@ -0,0 +1,16 @@
...
+Prisma will download the `linux-arm64-openssl-1.1.x` binary target, which depends on `libc`, which is not supported on Alpine by default.
+Running `prisma -v` will fail with:
+
+```sh
+Error: Unable to require(`/usr/src/app/node_modules/@prisma/engines/libquery_engine-linux-arm64-openssl-3.0.x.so.node`)
+  Error loading shared library ld-linux-aarch64.so.1: No such file or directory (needed by /usr/src/app/node_modules/@prisma/engines/libquery_engine-linux-arm64-openssl-1.1.x.so.node)
```
janpio, line 15: Ah nice how this has a different error message than the similar `with-libc` setup above 👍

docker-unsupported/alpine-3.17-arm64-openssl-3.0.x-with-libc/README.md
```diff
@@ -0,0 +1,16 @@
+# OS support: fail-alpine-3.17-arm64-openssl-3.0.x
+
+- Base Docker image: `node:lts-alpine3.17`
+- OS: Linux Alpine 3.17
+- Arch: arm64 (aarch64)
+- OpenSSL location: `/lib/libssl.so.3`
+- Binary target: `linux-arm64-openssl-3.0.x`
+
+Prisma doesn't support the `arm64` architecture for Linux Alpine.
```
janpio: Link?

docker-unsupported/alpine-3.17-arm64-openssl-3.0.x/README.md
```diff
@@ -0,0 +1,16 @@
+# OS support: alpine-3.17-arm64-openssl-3.0.x
+
+- Base Docker image: `node:lts-alpine3.17`
+- OS: Linux Alpine 3.17
+- Arch: arm64 (aarch64)
+- OpenSSL location: `/lib/libssl.so.3`
+- Binary target: `linux-arm64-openssl-3.0.x`
+
+Prisma doesn't support the `arm64` architecture for Linux Alpine.
```
janpio: Link?

docker-unsupported/alpine-3.16-arm64-openssl-1.1.x-with-libc/README.md
jkomyno: https://github.com/prisma/prisma/issues/8478

docker/distroless-bullseye-amd64-openssl-1.1.x/README.md
```diff
@@ -0,0 +1,10 @@
+# OS support: distroless-bullseye-amd64-openssl-1.1.x
+
+- Base Docker image: `gcr.io/distroless/nodejs18-debian11:debug`
+- OS: Linux Debian Bullseye, Distroless
+- Arch: amd64 (x86_64)
+- OpenSSL location: `/usr/lib/x86_64-linux-gnu/libssl.so.1.1`
+- Binary target: `debian-openssl-1.1.x`
+
+Distroless images are a set of minimal images containing only your application and its runtime dependencies. They do not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution.
+We had to copy the `libz.so.1` file from the `node:18-bullseye-slim` image (which uses the same Linux Debian version and arch as the distroless image) to the distroless image to make Prisma work.
```
janpio: We are not actually running this test. Is that because if would fail, or because it does not matter right now?

janpio (approved):

docker-unsupported/alpine-3.16-arm64-openssl-1.1.x-with-libc/README.md
jkomyno: Added this to the `alpine` tests in `docker-unsupported`

docker/distroless-bullseye-amd64-openssl-1.1.x/README.md
jkomyno: As answered privately, the latter

jkomyno:
Note: `test / platforms-serverless-vercel` is failing due to https://github.com/prisma/prisma/pull/17241

PR #3371, opened 2023-01-12
feat: add almalinux latest amd64 and arm64

jkomyno:
Add Docker tests:
- [x] almalinux-latest-amd64-openssl-1.1.x
- [x] almalinux-latest-arm64-openssl-1.1.x

As usual, only the `Dockerfile`s,  and `README.md` were modified, as well as the Docker image name in `run.sh` and the package name in `package.json`. The rest is copy-pasted by other Docker tests.

This just happened while writing https://github.com/prisma/prisma/issues/14042#issuecomment-1380692643

>On December 8, 2020, [Red Hat](https://en.wikipedia.org/wiki/Red_Hat) announced that development of [CentOS Linux](https://en.wikipedia.org/wiki/CentOS_Linux), a free-of-cost [downstream](https://en.wikipedia.org/wiki/Downstream_(software_development)) fork of the commercial Red Hat Enterprise Linux (RHEL), would be discontinued and its official support would be cut short to focus on CentOS Stream, a stable LTS release without minor releases officially used by Red Hat to preview what is intended for inclusion in updates to RHEL.[[4]](https://en.wikipedia.org/wiki/AlmaLinux#cite_note-4)[[5]](https://en.wikipedia.org/wiki/AlmaLinux#cite_note-5)[[6]](https://en.wikipedia.org/wiki/AlmaLinux#cite_note-6)
>In response, CloudLinux – which maintains its own commercial Linux distribution, [CloudLinux OS](https://en.wikipedia.org/wiki/CloudLinux_OS) – created AlmaLinux to provide a community-supported spiritual successor to CentOS Linux, aiming for binary-compatibility with the current version of RHEL.[[7]](https://en.wikipedia.org/wiki/AlmaLinux#cite_note-FAQ-7)
https://en.wikipedia.org/wiki/AlmaLinux


Jolg42 (approved):
This looks good but add 4 more tests, it feels like we're adding quite a good amount these days, hopefully the DX with the CI looks ok for now (it's fast enough)

Maybe we should add this to https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#binarytargets-options (let's check that in Slack with others)

janpio:
Part of the higher level engine OS support project is to also create a new "supported OS" page in docs, there it would definitely fit.

PR #3373, opened 2023-01-13
feat: adapt tests to expect oldest `libssl` version available (after #17241), fix cache errors

jkomyno:
This follows [this comment](https://github.com/prisma/prisma/pull/17241#issuecomment-1380711117) in https://github.com/prisma/prisma/pull/17241. 

- [x] update `EXPECTED_PRISMA_TARGET_PLATFORM` for `docker/alpine:3.17` images to expect `openssl-1.1.x` rather than `openssl-3.0.x` when `libssl-compat` is installed
- [x] update `EXPECTED_PRISMA_TARGET_PLATFORM` for `docker/alpine:3.17` images to expect `openssl-3.0.x` rather than `openssl-3.0.x` by default
- [x] update `platform-serverless gcp/firebase` to expect `openssl-1.0.x` rather than `openssl-1.1.x`. `prisma` now prefers the oldest `libssl` version available due to Vercel
- [x] fix Vercel tests after the discoveries in https://github.com/prisma/prisma/issues/17264#issuecomment-1383725772
- [x] fix a random cache error on `almalinux` via https://stackoverflow.com/a/60585660/6174476 

**Note**: I recommend reviewing this PR by excluding the changes to `yarn.lock` files. You can use [these filters](https://github.com/prisma/ecosystem-tests/pull/3373/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.prisma&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension&show-viewed-files=true).


jkomyno:
Note: `test / platform-serverless (gcp-functions, binary, ubuntu-20.04)` is currently failing due to resource exhaustion:

```text
Maximum number of functions in region us-central1 in project *** has been reached
```

See [Github Actions logs](https://github.com/prisma/ecosystem-tests/actions/runs/3923129260/jobs/6706540017).

docker/alpine-3.17-amd64-openssl-3.0.x/Dockerfile
```diff
@@ -6,11 +6,11 @@ WORKDIR /usr/src/app
 COPY --from=app . ./
 COPY --from=utils . ./
 
-RUN apk update && apk add openssl1.1-compat
+RUN apk update
 
-# Both OpenSSL 1.1 and OpenSSL 3 must be present after installing openssl1.1-compat
-RUN if [ ! -e /lib/libssl.so.1.1 ]; then echo 'OpenSSL 1.1 not found in /lib/libssl.so.1.1'; exit 1; fi
+# Only OpenSSL 3 is expected to be on this system
```
jkomyno, line 11: .. as we've removed `openssl1.1-compat`, which instead still exists in the new `alpine-3.17-amd64-openssl-1.1.x` test

Jolg42: I guess you meant `alpine-3.17-amd64-openssl-1.1.x`?

platforms-serverless-vercel/vercel-cli/run.sh
```diff
@@ -8,18 +8,21 @@ yarn
...
 
-  yarn -s vercel --token=$VERCEL_TOKEN --build-env DEBUG="prisma:*" --build-env PRISMA_CLIENT_ENGINE_TYPE="binary" --prod --scope=$VERCEL_ORG_ID --confirm --force 1> deployment-url.txt
+  yarn -s vercel --token=$VERCEL_TOKEN --build-env PRISMA_CLIENT_ENGINE_TYPE="binary" --prod --scope=$VERCEL_ORG_ID --confirm --force 1> deployment-url.txt
 else
  echo "Library (Default)"
   export VERCEL_PROJECT_ID=$VERCEL_API_PROJECT_ID
   echo "VERCEL_PROJECT_ID: $VERCEL_PROJECT_ID"
 
-  yarn -s vercel --token=$VERCEL_TOKEN --build-env DEBUG="prisma:*" --build-env PRISMA_CLIENT_ENGINE_TYPE="library" --prod --scope=$VERCEL_ORG_ID --confirm --force 1> deployment-url.txt
+  yarn -s vercel --token=$VERCEL_TOKEN --build-env PRISMA_CLIENT_ENGINE_TYPE="library" --prod --scope=$VERCEL_ORG_ID --confirm --force 1> deployment-url.txt
```
Jolg42, line 25: Should we try `DEBUG=prisma:get-platform` instead?

platforms-serverless/firebase-functions/test.sh
```diff
@@ -7,9 +7,9 @@ url="https://us-central1-prisma-e2e-tests-265911.cloudfunctions.net/$func"
 prisma_version="$(cat ../../.github/prisma-version.txt)"
 
 if [ "$PRISMA_CLIENT_ENGINE_TYPE" == "binary" ]; then
-  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","package.json","query-engine-debian-openssl-1.1.x","schema.prisma"]'
+  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","package.json","query-engine-debian-openssl-1.0.x","query-engine-debian-openssl-1.1.x","schema.prisma"]'
 else
-  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-debian-openssl-1.1.x.so.node","package.json","schema.prisma"]'
+  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-debian-openssl-1.0.x.so.node","libquery_engine-debian-openssl-1.1.x","package.json","schema.prisma"]'
```
janpio:
```suggestion
  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-debian-openssl-1.0.x.so.node","libquery_engine-debian-openssl-1.1.x.so.node","package.json","schema.prisma"]'
```

janpio: Should this really include both now?

platforms-serverless-vercel/vercel-cli/run.sh
jkomyno: The `DEBUG` env var was only useful in the discovery phase. No need to add it back unless we have other problems with Vercel at a certain point in the future

platforms-serverless/firebase-functions/test.sh
jkomyno: This is because `binaryTargets = ["native", "debian-openssl-1.1.x"]` was defined in the original schemas (see [binary](https://github.com/prisma/ecosystem-tests/blob/982fb472e74c00d8ae42148262ef115ad61122a8/platforms-serverless/firebase-functions/functions/prisma/schema-with-binary.prisma#L3), [node-api](https://github.com/prisma/ecosystem-tests/blob/982fb472e74c00d8ae42148262ef115ad61122a8/platforms-serverless/firebase-functions/functions/prisma/schema-with-node-api.prisma#L3)). I bet if I remove that `binaryTargets` line, we'd only see one `libquery` engine in the `files` expectation in `test.sh`

Jolg42: In logs both are present indeed, that's new and unexpected to me

Jolg42: For firebase and Google Cloud functions (Using Ubuntu 18.04 image like in our setup) in general the expected OpenSSL version is `openssl 	1.1.1-1ubuntu2.1~18.04.19` so `debian-openssl-1.1.x` which means for the QR `query-engine-debian-openssl-1.1.x`
https://cloud.google.com/functions/docs/reference/system-packages#ubuntu_1804

This mentions that Ubuntu 18.04 is used for `nodejs14` runtime which we use in these tests
https://cloud.google.com/functions/docs/concepts/execution-environment

It's unexpected that `1.0.x` shows up here to me

docker/alpine-3.17-amd64-openssl-3.0.x/Dockerfile
jkomyno: I did, updated the comment.

platforms-serverless/firebase-functions/test.sh
janpio: So the native detection changed from also detecting `debian-openssl-1.1.x` to now `debian-openssl-1.0.x`?

aqrln: @Jolg42 this link says both OpenSSL 1.0 (1.0.2n-1ubuntu5.10) and OpenSSL 1.1 (1.1.1-1ubuntu2.1~18.04.19) are installed.
image

`openssl` package is OpenSSL CLI, and it indeed is only present for OpenSSL 1.1, while OpenSSL 1.0 exists only as a library on that system, but the library is what's important to us.

aqrln: @janpio we now prefer the oldest among the installed versions since https://github.com/prisma/prisma/pull/17241

janpio: This should fix the tests @jkomyno

janpio: Ok, then this change makes sense. Why do we even have `binaryTargets` in this test? Shouldn't we be able to remove it and update the test expectations to just the newly selected one and it should work?

jkomyno: I second what @aqrln has written. @janpio, the CI is green now. May somebody please approve?

janpio:
Can you update the PR title to tell something about changes and the PR description to reflect the actual diff? (The diff shows another docker image for example). Happy to review then again.

platforms-serverless/firebase-functions/test.sh
Jolg42: Thanks, @aqrln that makes sense!

.github/workflows/test.yaml
```diff
@@ -186,6 +186,7 @@ jobs:
           - almalinux-latest-amd64-openssl-1.1.x
           - almalinux-latest-arm64-openssl-1.1.x
           - alpine-3.16-amd64-openssl-1.1.x
+          - alpine-3.17-amd64-openssl-1.1.x
```
janpio, line 189: Unrelated: I think it would be a good idea to add the compat stuff (or manualy installation etc) in the name here so one knows where this openssl version comes from for each - makes it easier to understand that is happening. Follow up PR of course.

platforms-serverless-vercel/vercel-cli/run.sh
janpio: Maybe add it as a comment? It would not be super obvious to me how to add that when looking just at the file - a comment would help.

janpio (approved):

platforms-serverless/firebase-functions/functions/prisma/schema-with-binary.prisma
```diff
@@ -1,6 +1,5 @@
 generator client {
   provider      = "prisma-client-js"
-  binaryTargets = ["native", "debian-openssl-1.1.x"]
   engineType = "binary"
```
Jolg42, line 3: Nice! Now the schema is the same as Google Cloud Functions 🙌🏼

Jolg42 (approved):
This looks great! 🙌🏼

.github/workflows/test.yaml
jkomyno: Ok, maybe something like `-with-libssl-compat` as a suffix

platforms-serverless-vercel/vercel-cli/run.sh
jkomyno: There's already a comment at the top of the file, lines 11-12:

```bash
# Note: be aware that Vercel truncates logs, so if you add something like `--build-env DEBUG="prisma:*"` plenty of logs will be missing.
# That will likely influence the "Postinstall hook" check below, which will result in the CI failing with error code `1`.
```

Am I missing something else? @janpio

aqrln (approved):

platforms-serverless/firebase-functions/test.sh
Jolg42: Note: binaryTargets are now removed https://github.com/prisma/ecosystem-tests/pull/3373#pullrequestreview-1250314429

PR #3374, opened 2023-01-13
test: try vercel-cli with `prisma@4.8.0` and `prisma@4.8.1`

jkomyno:
The following ecosystem tests are failing on `dev` and on https://github.com/prisma/ecosystem-tests/pull/3373:
- test / platforms-serverless-vercel (vercel-cli, library)
- test / platforms-serverless-vercel (vercel-cli, binary)

In particular, we're printing `Postinstall hook was NOT ADDED` and explicitly exiting with status code 1.
Maybe Vercel changed something and this check is no longer necessary?

However, these tests succeed on:
- `prisma@4.8.0`
- `prisma@4.8.1`
- `prisma@4.9.0-integration-revert-openssl-investigation-follow-up.1`

and fail from `prisma@4.9.0-integration-revert-openssl-investigation-follow-up.2`.

`prisma@4.9.0-integration-try-vercel-libssl-things.1`, `prisma@4.9.0-integration-try-vercel-libssl-things.2` also failed, although there wasn't any reasonable explanation for that.
However, Vercel truncates log and only retains the latest 4KB of them, and using `DEBUG` buid-time environment variables (which were useful for understanding that `prisma` used to return the `rhel-openssl-1.0.x` platform on Vercel) resulted in large logs. In fact, these logs were so large that the `prisma generate || true` text printed by Prisma's postinstall hook wasn't captured by our `run.sh` script, resulting in the `Postinstall hook was NOT ADDED` line being displayed and the CI run to fail with exit code `1`.


jkomyno:
Closing this PR, as this was only useful for investigating failures on Vercel.

PR #3376, opened 2023-01-15
fix: align prisma version to latest dev

jkomyno:
Random find: judging from the package.json, it looks like we haven't tested recent versions of `prisma` in the `vercel-with-redwood` test suite in months. In reality, thanks to the `patch-package-json.js` utility, the tests would be run with the most recent Prisma version regardless (but with warnings, as can be seen in [these logs](https://github.com/prisma/ecosystem-tests/actions/runs/3923095329/jobs/6706492078#step:7:119)). This PR fixes that inconsistency.


millsp:
This is fixed in the new pnpm setup :)

PR #3385, opened 2023-01-18
feat: add support for archlinux on amd64

jkomyno:
I've tried running `archlinux` locally while reproducing https://github.com/prisma/prisma/discussions/17380, so I thought I may just push this here.

This PR adds the following new Docker tests for `archlinux-latest-amd64-openssl-3.0.x`.

I discovered that:
- `ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort` returns a non-`libssl` library, so `openssl` is run as a backup. Thus, in `@prisma/get-platform`, we should probably change `grep ssl` into `grep libssl` (as we may pick up libraries like libevent_openssl-2.1.so or libgnutls-openssl.so.27 rather than libssl.so*, e.g. in Arch Linux)
- While `archlinux` has both `libssl-1.0.x` and. `libssl-3.0.x` installed, `openssl version -v` resolves to `3.0.x`, which for the reasons above causes Prisma to download the `debian-openssl-3.0.x` flavor of the engines


Jolg42:
@jkomyno is that expected?
```
prisma:warn Prisma failed to detect the Linux distro in use, and may not work as expected. Defaulting to "debian".
```
https://github.com/prisma/ecosystem-tests/actions/runs/3948811899/jobs/6759291265#step:9:234

Jolg42 (approved):

jkomyno:
> @jkomyno is that expected?
> 
> ```
> prisma:warn Prisma failed to detect the Linux distro in use, and may not work as expected. Defaulting to "debian".
> ```
> 
> https://github.com/prisma/ecosystem-tests/actions/runs/3948811899/jobs/6759291265#step:9:234

Yes, that's exactly what the discussion linked in the PR description was about :)

PR #3415, opened 2023-01-30
feat: add docker tests for `distroless`

jkomyno:
Added back tests that were temporarily removed in https://github.com/prisma/ecosystem-tests/pull/3343.

This PR adds support for:
- [x] `docker/distroless-bullseye-amd64-openssl-1.1.x`
- [x] `docker-unsupported/distroless-bullseye-amd64-openssl-1.1.x-without-zlib`, which fails at runtime as expected due to the missing `zlib` library


jkomyno:
Note: apparently `PrismaClientInitializationError` doesn't result in a non-0 error code in `docker-unsupported/distroless-bullseye-amd64-openssl-1.1.x-without-zlib`

Issue #1629, opened 2021-04-16
Vercel equivalent to `netlify-github`/Git


janpio:
`netlify-github` is a Git project connected to the Netlify project that deploys on push. 
That is also possible on Vercel, so we should test that for completeness.

PR #3454, opened 2023-03-22
test(vercel, netlify): schema caching error messages

millsp:
[Relevant files to review](https://github.com/prisma/ecosystem-tests/pull/3454/files?file-filters%5B%5D=.js&file-filters%5B%5D=.md&file-filters%5B%5D=.prisma&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&file-filters%5B%5D=dotfile&show-viewed-files=true)

This PR ensures that the changes brought by https://github.com/prisma/prisma/pull/18437 work. The original commits on this PR were to reproduce the caching errors, but has now been repurposed to check the change. Please ignore all the other failing tests as they are not relevant or broken by this PR.

Closes https://github.com/prisma/client-planning/issues/296


platforms-serverless-vercel/vercel-with-nextjs-caching/script2.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# to check that manual command removes caching error
```
janpio, line 3: echo "running script.sh that manually runs 'prisma generate'"

platforms-serverless-vercel/vercel-with-nextjs-caching/script1.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+# do nothing here to trigger the caching error message
```
janpio, line 3:
```suggestion
echo "running script.sh that does nothing, to trigger the error"
# do nothing here to trigger the caching error message
```

platforms-serverless-vercel/vercel-with-nextjs-caching/script2.sh
janpio, line 1: Uh, ok, now I understand the test setup of `vercel-with-nextjs-caching` - it is `vercel deploy` and somehow `script.sh` is used as a "Build command" replacement via `package.json`. Correct?

Then 2 suggestions:
- Let's split the failing and successful build into 2 different projects - so it is easier to look at those builds and know what one is looking at. Right now I have no idea what I am looking at in the project build history in Vercel itself
- We should probably also have a `vercel-github` setup that uses the Git integration of Vercel instead of local `vercel deploy` as that is much more common with users. (That maybe implies that we _just_ test the successful case with `prisma generate` set into the build command via the web UI - unless we do multiple of those even (one via script and package.json as here, one via web UI))
	- Ah, I see we have a long time issue for that lack of Git(hub) based test: https://github.com/prisma/ecosystem-tests/issues/1629

platforms-serverless-vercel/vercel-with-nextjs-caching/README.md
janpio, line 1: We usually replace these with a description of the test project.

platforms-serverless/netlify-github-with-nextjs-caching/README.md
janpio, line 1: We usually replace these with a description of the test project.

platforms-serverless/netlify-github-with-nextjs-caching/prepare.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+set -eu
+
+export PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platforms-serverless netlify-github-with-nextjs-caching build'
+
+rm -fr .git # because caanot currently be done in finally.sh if failure happens
```
janpio:
```suggestion
rm -fr .git # because cannot currently be done in finally.sh if failure happens
```

janpio: What is this needed for?

platforms-serverless/netlify-github/finally.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-echo "TODO: Read Netlify logs (when available via https://github.com/netlify/cli/issues/119 )"
\ No newline at end of file
+echo "TODO: Read Netlify logs (when available via https://github.com/netlify/cli/issues/119)"
```
janpio:
```suggestion
# intentional space before closing bracket so clickabel link is valid in CI output - otherwise it is also linked and link is invalid
echo "TODO: Read Netlify logs (when available via https://github.com/netlify/cli/issues/119 )"
```

janpio:
When reading the test run like this:
```
$ jest index.test.js
 LOGGING RETRY ERRORS  simple query
 RETRY 1 

    FetchError: invalid json response body at https://1680045935854107570--netlify-github-with-nextjs-caching.netlify.app//api reason: Unexpected token N in JSON at position 0

      16 |
      17 |   const r = await fetch(endpoint + '/api')
    > 18 |   const data = await r.json()
         |                ^
      19 |   expect(data).toMatchObject(***
      20 |     ok: true,
      21 |   ***)

      at node_modules/node-fetch/lib/index.js:273:32
      at Object.<anonymous> (index.test.js:18:16)

 RETRY 2 

    FetchError: invalid json response body at https://1680045935854107570--netlify-github-with-nextjs-caching.netlify.app//api reason: Unexpected token N in JSON at position 0

      16 |
      17 |   const r = await fetch(endpoint + '/api')
    > 18 |   const data = await r.json()
         |                ^
      19 |   expect(data).toMatchObject(***
      20 |     ok: true,
      21 |   ***)

      at node_modules/node-fetch/lib/index.js:273:32
      at Object.<anonymous> (index.test.js:18:16)

PASS ./index.test.js (16.431 s)
  ✓ simple query (10605 ms)
```
https://github.com/prisma/ecosystem-tests/actions/runs/4548560487/jobs/8019747237#step:7:370

Does this mean that try 1 and 2 failed and then try 3 succeeded as intended?

platforms-serverless-vercel/vercel-with-nextjs-caching/script2.sh
millsp: > Let's split the failing and successful build into 2 different projects - so it is easier to look at those builds and know what one is looking at. Right now I have no idea what I am looking at in the project build history in Vercel itself

I'd argue that this currently represents more the "real" flow of a user:
- First we create a project at the given platform, in a fresh state, then the dependency gets cached.
- We then access the deployed code and find the error, which is recorded in jest and in the logs.
- We fix the error as suggested by the documentation, we add a postinstall hook in our package.json.
- Then we deploy again, with the cache still being there, we don't clear or force deploy this time.
- We then access the app, and see that everything works as it should.

I expect that to be more or less the user journey when they see the message for the first time.

As for the logs, the best is to run the script and click on the links when they appear and watch the Vercel console.

> We should probably also have a vercel-github setup that uses the Git integration of Vercel instead of local vercel deploy as that is much more common with users.

I agree, do you think we need to work on this right now or can we schedule an issue and work on it soon?

platforms-serverless-vercel/vercel-with-nextjs-caching/script2.sh
millsp: I think the log output of that would be swallowed by the package manager, or could show up in the vercel logs only.

platforms-serverless/netlify-github-with-nextjs-caching/prepare.sh
millsp: When there's a failure and the run is fully retried, there is no way to properly clean it up because `finally.sh` is not called. So the git commands will actually fail. To work around that we always delete the git files. The actual problem might actually be that we don't reset the test folder properly, so maybe we should properly reset before running a test so it can be re-run without keeping the previous state. Something along these lines:

```sh
git reset HEAD --hard && git clean -f -x && <run test command>
```

millsp:
> Does this mean that try 1 and 2 failed and then try 3 succeeded as intended?

Yes, jest does that

platforms-serverless-vercel/vercel-with-nextjs-caching/script2.sh
janpio: Issue.

millsp: Added it as a tech debt project for more visibility.

platforms-serverless/netlify-github-with-nextjs-caching/prepare.sh
millsp: Created a draft issue as a reminder

janpio (approved):

PR #3461, opened 2023-03-28
test(docker): add centos 7 tests

aqrln:


jkomyno:
I see AWS Graviton failing and old others, but that seems unrelated to this PR. Is there a quick fix for it that we can merge in `dev` before this PR, just to see all tests passing?

Jolg42:
@aqrln Just found this randomly, what is the status on this? (Why is it needed? Asking because description is empty)

aqrln:
@Jolg42 I don't remember for sure either 😅
I vaguely remember that it prompted by some issue, maybe some user reported Prisma not working on CentOS?

It probably makes sense to add this regardless of the original motivation, since it's the oldest platform we support and I don't think we have explicit tests for it.

aqrln:
It looks like a rebase went wrong on this branch, I can fix it later

PR #3463, opened 2023-03-29
chore: use pnpm instead of yarn

millsp:
[Relevant files for the review](https://github.com/prisma/ecosystem-tests/pull/3463/files?file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.sh&file-filters%5B%5D=.ts&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&file-filters%5B%5D=dotfile&show-deleted-files=true&show-viewed-files=true#diff-f2fb5137a88c12ca5c4cd632943c64616acbe8e340fb91f060317abfce50b49e)

Modernize tooling.


janpio:
Macstadium does not have `pnpm` installed yet:
```
uname arm64
arm true
m1 true
zsh:9: command not found: pnpm
Connection to 207.254.29.83 closed.
```

dataproxy/vercel-cli-serverless-functions/run.sh
```diff
@@ -22,17 +22,9 @@ echo "Deployed to ${DEPLOYED_URL}"
...
 echo "${OUTPUT}"
 
-# Check the Vercel Build Logs for the postinstal hook"
-if echo "${OUTPUT}" | grep -q 'prisma generate || true'; then
-  echo 'Postinstall hook was added'
-else
-  echo "Postinstall hook was NOT ADDED"
-  exit 1
-fi
-
```
janpio, line 35: unrelated

frameworks/sveltekit/package.json
janpio, line 1: Project regenerated or why all these "unrelated" changes?

dataproxy/vercel-cli-serverless-functions/run.sh
millsp: It does not work in some cases with pnpm it seems

frameworks/sveltekit/package.json
millsp: Project was outdated for a long long time and pnpm did not deal weel with the way deps were setup, so I "just" upgraded to the latest Svelte

jkomyno:
@millsp I see this is in review, but there are plenty of lock breaking changes due to a new integration version of Prisma being released. Could you please fix them and ping us right away (possibly on Slack) for the review? Thanks

aqrln:
@jkomyno 

> @millsp I see this is in review, but there are plenty of lock breaking changes due to a new integration version of Prisma being released. Could you please fix them and ping us right away (possibly on Slack) for the review? Thanks

I think that should happen independently of the code review, and probably even only after getting the approvals tbh. Otherwise it's just extra work for no good reason — those conflicts may happen again before people even start reviewing.

.github/scripts/check-engines-cli.sh
```diff
@@ -88,17 +88,17 @@ elif [ $CLI_QUERY_ENGINE_TYPE == "<dataproxy>" ]; then
   echo "DataProxy: Enabled"
 else
   echo "❌ CLI_QUERY_ENGINE_TYPE was not set"
-  exit 1
+  #exit 1
```
aqrln: Is this intentional? Can we remove it instead if it's not necessary anymore?

.github/workflows/optional-test.yaml
```diff
@@ -88,17 +90,18 @@ jobs:
...
-          # Only cache the root node_modules
-          cache-dependency-path: 'yarn.lock'
+          cache: 'pnpm'
       - name: report failure to slack
         run: bash .github/scripts/slack-workflow-status.sh "(Optional tests) :x:"
 
   community-generators:
     needs: [detect_jobs_to_run]
-    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['community-generators'] == true }}
+    if: ${{ false && fromJson(needs.detect_jobs_to_run.outputs.jobs)['community-generators'] == true }}
```
aqrln: Is this still needed?

.github/workflows/optional-test.yaml
```diff
@@ -141,7 +146,7 @@ jobs:
 
   databases:
     needs: [detect_jobs_to_run]
-    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['databases'] == true }}
+    if: ${{ false && fromJson(needs.detect_jobs_to_run.outputs.jobs)['databases'] == true }}
```
aqrln: same, probably needs to be re-enabled now

.github/scripts/check-engines-cli.sh
millsp: Thanks for catching this. Forgot to port the logic to be able to resolve pnpm locations in a way that works universally across all package managers.

.github/scripts/check-engines-cli.sh
```diff
@@ -52,36 +52,32 @@ esac
 echo "Assumed OS: $os_name"
 echo "CLI_QUERY_ENGINE_TYPE == $CLI_QUERY_ENGINE_TYPE"
 
+ENGINES_PACKAGE=$(node -e "console.log(path.dirname(require.resolve('@prisma/engines/package.json', {paths: [path.dirname(require.resolve('prisma/package.json'))]})))")
```
jkomyno, line 55: This looks much nicer than before! Can we also echo it, just in case for the future?

jkomyno (approved):
Nice job, approved with a minor comment. The only actual failing test is azure-serverless, which seems to be failing unrelated to the content of this PR

.github/scripts/check-engines-cli.sh
```diff
@@ -52,36 +52,32 @@ esac
...
+
 if [ $CLI_QUERY_ENGINE_TYPE == "binary" ]; then
   echo "Binary: Enabled"
   case $os_name in
     linux)
-      qe_location="node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x"
-      qe_location2="node_modules/prisma/node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x"
+      qe_location="$ENGINES_PACKAGE/query-engine-debian-openssl-1.1.x"
```
janpio, line 61: These are two different locations that are usually both with a QE. Why only check one now?

.github/scripts/check-engines-cli.sh
```diff
@@ -92,13 +88,8 @@ else
 fi
 
 
-echo "--- yarn prisma -v ---"
-yarn -s prisma -v
-echo "--- ls -lh node_modules/@prisma/engines/ ---"
-ls -lh node_modules/@prisma/engines/
-echo "--- ls -lh node_modules/prisma/ ---"
-ls -lh node_modules/prisma/
```
janpio, line 100: This used to be useful to get an overview of the files. 
Not possible any more with pnpms placing of node_modules I guess?

.github/scripts/check-engines-cli.sh
millsp: I thought the one was a fallback because `@prisma/engines` is always a dependency of `prisma` and is never installed on it's own. For that reason, I opted for looking it up via it's parent dependency instead of via it's potential direct path. As I understood, the one is a fallback for the other.

.github/scripts/check-engines-cli.sh
millsp: I can restore it

.github/scripts/check-engines-cli.sh
janpio: Probably correct. Line 107 can be updated as well then to not check the second path any more.

janpio: (Although I wonder a bit why the fallback was necessary... if it was.)

.github/scripts/check-engines-cli.sh
janpio: Not really necessary I think. Maybe commented out for when we need to debug things in this script.

.github/workflows/optional-test.yaml
```diff
@@ -108,7 +111,6 @@ jobs:
     runs-on: ubuntu-20.04
 
     env:
-      SKIP_PRISMA_VERSION_CHECK: true # see https://github.com/MichalLytek/typegraphql-prisma/issues/31
```
janpio, line 111: I think that is still needed, as otherwise the check will fail for some versions until typegraphql-prisma is updated.

millsp: I moved it into the script directly as it was not obvious to me when running locally

.github/scripts/check-engines-cli.sh
millsp: I think the fallback would be useful for different package managers, as they nest and hoist packages differently. This version  for finding the package is more universal. I will remove the forgotten check in the if.

.github/workflows/test.yaml
```diff
@@ -281,6 +288,19 @@ jobs:
...
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: 'pnpm'
+          cache-dependency-path: ${{ github.job }}/${{ matrix.feature }}/pnpm-lock.yaml
+
+      - name: Install Dependencies
+        run: pnpm install
+
```
janpio, line 303: Do we need install here? Doesn#t that all happen in Docker?

.github/workflows/test.yaml
```diff
@@ -712,25 +771,26 @@ jobs:
...
+          node-version: 16
+          cache: ${{ contains(matrix.platform, 'codesandbox') && 'npm' || 'pnpm' }}
+          cache-dependency-path: ${{ github.job }}/${{ matrix.platform }}/${{ contains(matrix.platform, 'codesandbox') && 'package-lock.json' || 'pnpm-lock.yaml' }}
 
       - name: Define Client Engine Type to test
         run: echo "PRISMA_CLIENT_ENGINE_TYPE=${{ matrix.clientEngine }}" >> $GITHUB_ENV
 
       - name: Install Dependencies
-        run: yarn install
+        run: ${{ contains(matrix.platform, 'codesandbox') && 'npm install' || 'pnpm install' }}
```
janpio: pnpm does not work on codesandbox?

.github/workflows/test.yaml
```diff
@@ -800,18 +860,21 @@ jobs:
...
+
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
-          cache: 'yarn'
-          # Only cache the root node_modules
-          cache-dependency-path: 'yarn.lock'
+          node-version: 16
+          cache: ${{ contains(matrix.platform, 'azure') && 'npm' || 'pnpm' }}
+          cache-dependency-path: ${{ github.job }}/${{ matrix.platform }}/${{ contains(matrix.platform, 'azure') && 'package-lock.json' || 'pnpm-lock.yaml' }}
```
janpio: ?

.github/workflows/test.yaml
```diff
@@ -869,18 +932,21 @@ jobs:
...
+          node-version: 16
+          cache: ${{ contains(matrix.platform, 'redwood') && 'yarn' || 'pnpm' }}
+          cache-dependency-path: ${{ github.job }}/${{ matrix.platform }}/${{ contains(matrix.platform, 'redwood') && 'yarn.lock' || 'pnpm-lock.yaml' }}
 
       - name: Define Client Engine Type to test
         run: echo "PRISMA_CLIENT_ENGINE_TYPE=${{ matrix.clientEngine }}" >> $GITHUB_ENV
 
       - name: Install Dependencies
-        run: yarn install
+        run: ${{ contains(matrix.platform, 'redwood') && 'yarn' || 'pnpm install' }}
```
janpio: ?

platforms-serverless/firebase-functions/functions/pnpm-lock.yaml
```diff
@@ -0,0 +1,1625 @@
+lockfileVersion: 5.4
```
janpio: A bit weird that we have a package.json and lockfile in the functions folder itself.

.github/workflows/test.yaml
millsp: Codesandbox automatically install npm or yarn but not pnpm

.github/workflows/test.yaml
millsp: Yes, nice catch

.github/workflows/test.yaml
millsp: There's not ternaries in workflow syntax so this is conditional cache selection

millsp: It improved job speed

.github/workflows/test.yaml
millsp: Similar to https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1159861542

millsp: Redwood is stuck on yarn so we have to run it with it

platforms-serverless/firebase-functions/functions/pnpm-lock.yaml
millsp: Good catch, will add an exclusion (firebase does not support pnpm)

millsp: There's only a handful of these. Such projects will have an exclusion via `.npmrc` (if you want to look them up) to tell pnpm not to generate a lockfile.

platforms-serverless/firebase-functions/prepare.sh
```diff
@@ -2,4 +2,4 @@
 
 set -eux
 
-npm install -g firebase-tools@9.14.0
+pnpm install -g firebase-tools@11.25.2
```
janpio: we should probably try to do that via package.json so this can automatically be updated

platforms-serverless/lambda/update-code.sh
```diff
@@ -1,10 +1,10 @@
-#!/bin.sh
```
janpio, line 1: Important change? If so, we should also search other files and make sure they all have that.

platforms-serverless-vercel/vercel-node-builder/test.sh
```diff
@@ -4,9 +4,9 @@ set -eux
 DEPLOYED_URL=$( tail -n 1 deployment-url.txt )
 
 if [ "$PRISMA_CLIENT_ENGINE_TYPE" == "binary" ]; then
-  files=',"files":["index.js","package.json","query-engine-rhel-openssl-1.0.x","schema.prisma"]'
+  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","schema.prisma"]'
 else
-  files=',"files":["index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","schema.prisma"]'
+  files=',"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","schema.prisma"]'
```
janpio, line 9: Why the different file list?

platforms/aws-graviton/code/index.test.js
```diff
@@ -22,13 +22,13 @@ describe('Prisma', () => {
...
+    const generatedClientDir = path.dirname(require.resolve('.prisma/client', {
+      paths: [path.dirname(require.resolve('@prisma/client'))],
+    }))
+    const files = fs.readdirSync(generatedClientDir)
     if (process.env.PRISMA_CLIENT_ENGINE_TYPE !== 'binary') {
       expect(files).toMatchInlineSnapshot(`
 Array [
-  "deno",
-  "edge.d.ts",
-  "edge.js",
```
janpio, line 31: Why files removed?

platforms-serverless/lambda/update-code.sh
millsp: Nothing complained but seemed incorrect

millsp: Yes, there is another instance o this. Will fix.

platforms/codesandbox/package.json
```diff
@@ -4,21 +4,21 @@
...
-    "node-fetch": "^2.6.0",
-    "puppeteer": "^19.0.0"
+    "node-fetch": "^2.6.0"
   },
   "devDependencies": {
     "@types/node": "14.18.36",
     "@types/node-fetch": "2.6.2",
     "@types/puppeteer": "5.4.7",
     "prisma": "4.13.0-dev.19",
+    "puppeteer": "^19.0.0",
```
janpio, line 21: do we still need this at all?

platforms/m1-macstadium/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "m1-macstadium",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "devDependencies": {
+    "prisma": "4.13.0-dev.19"
+  },
+  "dependencies": {
+    "@prisma/client": "4.13.0-dev.19"
+  }
```
janpio: Where did dependencies come from before?
Sure this folder is/was not excluded from auto updating?

platforms-serverless/firebase-functions/functions/package.json
```diff
@@ -8,7 +8,7 @@
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "postinstall": "CI=1 && PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platforms/firebase-functions postinstall' && yarn prisma generate"
+    "gcp-build": "CI=1 && PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platforms/firebase-functions postinstall' && npx prisma generate"
```
janpio, line 11: Uh, platform specific build command?
Should probably document if it is _needed_ for anything.

Also:
```suggestion
    "gcp-build": "CI=1 && PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platforms/firebase-functions gcp-build' && npx prisma generate"
```

platforms-serverless/firebase-functions/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "devDependencies": {
+    "prisma": "4.13.0-dev.19"
+  },
+  "dependencies": {
+    "@prisma/client": "4.13.0-dev.19"
+  }
+}
```
janpio: Necessary although package not even used here, but just in "functions"?

platforms-serverless/lambda/index.ts
```diff
@@ -11,7 +11,10 @@ const measure_client = process.hrtime.bigint()
 export async function handler() {
   const fs = require('fs')
   const path = require('path')
-  const files = fs.readdirSync(path.dirname(require.resolve('.prisma/client')))
+  const generatedClientDir = path.dirname(require.resolve('.prisma/client', {
+    paths: [require.resolve('@prisma/client')]
+  }))
+  const files = fs.readdirSync(generatedClientDir)
```
janpio, line 17: Does that still find wherever Prisma CLinet was generated to?

platforms-serverless/netlify-github/package.json
```diff
@@ -9,8 +9,8 @@
     "ts-node": "10.9.1"
   },
   "scripts": {
-    "dev": "npx netlify dev",
-    "build": "PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platform netlify-github build' && yarn prisma generate"
+    "dev": "pnpm netlify dev",
+    "build": "PRISMA_TELEMETRY_INFORMATION='ecosystem-tests platform netlify-github build' && pnpm prisma generate"
```
janpio, line 13: Third alterantive to `postinstall` or setting build script via UI I guess?

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
```diff
@@ -1,7 +1,7 @@
 generator client {
+  output = "client"
```
janpio: Isn't that buggy right now?

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
```diff
@@ -1,7 +1,7 @@
...
   provider      = "prisma-client-js"
-  // second binaryTarget `rhel` is required for deploying to AWS Lambda
-  binaryTargets = ["native", "rhel-openssl-1.0.x"]
```
janpio: Removing this makes it not work on local Windows/Mac/Ubuntu any more etc.

platforms-serverless/serverless-framework-lambda/index.ts
```diff
@@ -1,4 +1,5 @@
-import { PrismaClient, Prisma } from '@prisma/client'
+import { PrismaClient, Prisma } from './prisma/client'
```
janpio: Best practive of importing from `@prisma/client` not possible on Lambda via Serverless?

platforms-serverless/serverless-framework-lambda/serverless.yml
```diff
@@ -19,7 +19,8 @@ functions:
 # only include the Prisma binary required on AWS Lambda while packaging
 package:
   patterns:
-    - '!node_modules/.prisma/client/query-engine-*'
-    - '!node_modules/.prisma/client/libquery_engine-*'
-    - 'node_modules/.prisma/client/libquery_engine-rhel-openssl-1.0.x.so.node'
-    - 'node_modules/.prisma/client/query-engine-rhel-openssl-1.0.x'
+    - '!node_modules/**'
```
janpio: Uh completely excluding node_modules? Why?

platforms-serverless/serverless-framework-lambda/serverless.yml
```diff
@@ -19,7 +19,8 @@ functions:
...
+    - '!prisma/client/query-engine-*'
+    - '!prisma/client/libquery_engine-*'
+    - 'prisma/client/libquery_engine-rhel-openssl-1.0.x.so.node'
+    - 'prisma/client/query-engine-rhel-openssl-1.0.x'
```
janpio: I think this is from a best practice article we have somewhere.
If we change this here, we should also update the article in docs.

platforms-serverless/serverless-framework-lambda/test.ts
```diff
@@ -10,8 +10,8 @@ async function main() {
 
   const actual = (data.$response.data as any).Payload
   const binaryString = process.env.PRISMA_CLIENT_ENGINE_TYPE === 'binary'
-    ? `,"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","schema.prisma"]`
-    : `,"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","schema.prisma"]`
+    ? `,"files":["index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","runtime","schema.prisma"]`
+    : `,"files":["index-browser.js","index.d.ts","index.js","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","runtime","schema.prisma"]`
```
janpio: Why the difference here?

platforms-serverless-vercel/vercel-node-builder/index.js
```diff
@@ -39,7 +40,10 @@ app.get('/', async (req, res) => {
 
   // list all files in node_modules/.prisma/client
   const fs = require('fs')
-  const files = fs.readdirSync(process.env.LAMBDA_TASK_ROOT + "/node_modules/.prisma/client")
+  const generatedClientDir = path.dirname(require.resolve('.prisma/client', {
+    paths: [path.dirname(require.resolve('@prisma/client'))]
+  }))
+  const files = fs.readdirSync(generatedClientDir)
```
janpio, line 46: Ok, we have that so often now that I have to ask:
What exactly does this do, why and how?

platforms-serverless-vercel/vercel-with-redwood/web/package.json
```diff
@@ -13,9 +13,11 @@
     ]
   },
   "dependencies": {
+    "@prisma/client": "4.13.0-dev.19",
     "@redwoodjs/forms": "3.8.0",
     "@redwoodjs/router": "3.8.0",
     "@redwoodjs/web": "3.8.0",
+    "prisma": "4.13.0-dev.19",
```
janpio: This project does not need Prisma. Do we have to install it here?

platforms-serverless-vercel/vercel-with-redwood/api/package.json
```diff
@@ -3,8 +3,10 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "@prisma/client": "4.13.0-dev.19",
     "@redwoodjs/api": "3.8.0",
-    "@redwoodjs/graphql-server": "3.8.0"
+    "@redwoodjs/graphql-server": "3.8.0",
+    "prisma": "4.13.0-dev.19"
```
janpio: This additiona vs. the "resolutions" a few lines below?
One can go I guess?

platforms-serverless-vercel/vercel-with-redwood/package.json
```diff
@@ -30,5 +30,9 @@
   },
   "scripts": {
     "test": "yarn jest ./index.test.js"
+  },
+  "dependencies": {
+    "@prisma/client": "4.13.0-dev.19",
+    "prisma": "4.13.0-dev.19"
```
janpio: Not really needed on this level, right?

README.md
```diff
@@ -29,7 +29,7 @@ Test are then triggered via `test.sh`.
 
 Any clean up or logging work can then be done in `finally.sh` which is executed in all cases, even when the tests fail.
 
-**Note:** You need to use [yarn](https://yarnpkg.com) as it's used for bumping dependencies; i.e. run `yarn install` as a first step in your `run.sh` script.
+**Note:** You need to use [yarn](https://yarnpkg.com) as it's used for bumping dependencies; i.e. run `pnpm install` as a first step in your `run.sh` script.
```
janpio: update to mention pnpm, not yarn

.github/workflows/optional-test.yaml
janpio: Good point!

.github/workflows/upgrade-branch-to-version.yaml
```diff
@@ -1,51 +0,0 @@
-name: upgrade-branch-to-version
```
janpio, line 1: Why completely delete this? Did it not actually work?

package.json
```diff
@@ -14,6 +14,9 @@
     "typescript": "4.9.3"
   },
   "scripts": {
-    "fetch-retry": "yarn ts-node utils/fetch-retry-and-confirm-version.ts"
+    "fetch-retry": "pnpm exec ts-node utils/fetch-retry-and-confirm-version.ts",
+    "update-ci": "scripts/update-ci.sh",
+    "update-all": "scripts/update-all.sh",
+    "update-project": "scripts/update-project.sh"
```
janpio, line 20: Not used anywhere?

platforms-serverless/firebase-functions/functions/package.json
millsp: Yep, it has changed in the new versions, same for firebase.

platforms-serverless-vercel/vercel-node-builder/test.sh
millsp: That's a good question, and I don't have a good answer to that. I added this as a thing to follow up on in the document I shared, sign this is not well tested. This is the expected list, not having these is likely a bug.

platforms/aws-graviton/code/index.test.js
millsp: https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1160126091

millsp: Similar situation, just the other way around. I think we have bugs in both npm/yarn and pnpm.

platforms/codesandbox/package.json
millsp: I think so since we use the new sandboxes, I'll give it a try. As I understood, pupeteer was just used to wake up the sandbox.

platforms/m1-macstadium/package.json
millsp: The folders need to have a package.json for the script to find them. That said, we can remove the dependencies.

platforms-serverless/firebase-functions/package.json
millsp: Same as https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1160128518

platforms-serverless/lambda/index.ts
millsp: Yes, it works for pnpm and is compatible with all package managers

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
millsp: Serverless framework does not understand pnpm, I could switch the project back to npm (not yarn please).

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
millsp: Yes, I think that needs rework (I was lazy), somehow the excluded engine is still included. I take a look.

platforms-serverless/serverless-framework-lambda/index.ts
millsp: Related to https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1160130018
We generate in a custom folder

platforms-serverless/serverless-framework-lambda/serverless.yml
millsp: Related to https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1160130018
We generated in a  custom folder, so no need for node_modules

platforms-serverless/serverless-framework-lambda/test.ts
millsp: Custom outputs include runtime files, not the default output

scripts/update-all.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+NEW_VERSION=$(npm show prisma@$1 version)
+echo "$NEW_VERSION" > .github/prisma-version.txt
```
janpio, line 5: Isn't this called with an explicit version string in $1, and doing these two things will jsut return the same again?

scripts/update-all.sh
```diff
@@ -0,0 +1,7 @@
...
+
+pnpm -rc --parallel exec "$(pwd)/scripts/update-version.sh $NEW_VERSION"
+pnpm -rc --parallel exec "$(pwd)/scripts/update-locks.sh $NEW_VERSION"
```
janpio: This runs the script on all the projects/package.json it can find?

platforms-serverless-vercel/vercel-node-builder/index.js
millsp: This essentially helps to find the real location of the generated prisma client. Because pnpm plays with symlinks, hardlinks, and hoisting, it is able to artificially create isolated modules although they're all present in a global store. This is also why pnpm is considered a strict package manager, and that is good. This is also perfectly fine for node's resolution algorithm.

But that also means you cannot just grab the dependency at the root of your node_modules anymore, as they are strictly isolated. `.prisma/client` is always generated near `@prisma/client`, so this is what it means:
1. Look for `@prisma/client` and get that path
2. Look for `.prisma/client` from `@prisma/client`'s path

platforms-serverless-vercel/vercel-with-redwood/web/package.json
millsp: No, that's a mistake.

platforms-serverless-vercel/vercel-with-redwood/api/package.json
millsp: Yes, I'll clean that up

package.json
millsp: used via `pnpm update-ci` or `pnpm run update-ci`

scripts/update-all.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+NEW_VERSION=$(npm show prisma@$1 version)
+echo "$NEW_VERSION" > .github/prisma-version.txt
+
+pnpm -rc --parallel exec "$(pwd)/scripts/update-version.sh $NEW_VERSION"
```
janpio, line 8: Why string-replace of the version first in package.json, and then update the lockfile? Becuase it is faster?

scripts/update-ci.sh
```diff
@@ -0,0 +1,41 @@
...
+done
+
+echo "New version was found and will be used for the update."
+
+git fetch github "dev"
+git reset --hard "github/dev"
+
+pnpm run update-all "$NEW_VERSION"
+
+git commit -am "chore: sync, use $version"
```
janpio: $version is undefined I think

.github/workflows/upgrade-branch-to-version.yaml
millsp: The shell script was removed as it was the old version, but I can easily add this with a pnpm command

scripts/update-ci.sh
```diff
@@ -0,0 +1,41 @@
...
+  if [ $(($(date +%s) - $TIME_START)) -gt 240 ]
+  then
+    echo "Exited after no new version was found after 4 minutes."
+    exit 0
+  fi
+
+  NEW_VERSION=$(npm show prisma@$1 version)
+done
+
+echo "New version was found and will be used for the update."
```
janpio:
```suggestion
echo "New version $NEW_VERSION was found and will be used for the update."
```

scripts/update-ci.sh
```diff
@@ -0,0 +1,41 @@
...
+git reset --hard "github/$1"
+git checkout "github/$1"
+
+OLD_VERSION=$(cat .github/prisma-version.txt)
+NEW_VERSION=$OLD_VERSION
+TIME_START=$(date +%s)
+
+while [ "$NEW_VERSION" = "$OLD_VERSION" ]
+do
+  if [ $(($(date +%s) - $TIME_START)) -gt 240 ]
```
janpio, line 25:
```suggestion
  # GitHub Actions workflow is scheduled every 5 minutes, so exit here after 4 minutes
  if [ $(($(date +%s) - $TIME_START)) -gt 240 ]
```

scripts/update-all.sh
millsp: Yes

scripts/update-ci.sh
janpio:
```suggestion
git commit -am "chore: update to prisma@$NEW_VERSION"
```

scripts/update-all.sh
millsp: Yes, it's a small tweak for running `update-all` (eg `update-all dev`) locally while `update-ci` is meant just for ci.

scripts/update-all.sh
millsp: It is faster yes and also enabled `--parallel` which can run in any order, in which case, if you are in a workspace it won't matter if you ran `install`, it could be at the root or in a sub-project it won't a matter because versions are already updated everywhere (so mostly a monorepo handling thingy + performance)

scripts/update-ci.sh
millsp: Nice catch! Thanks

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-  fi
-  echo "$v (via Npm)"
-  
-  packages=$(find . -not -path "*/node_modules/*" -type f -name "package.json")
-  echo "$packages" | tr ' ' '\n' | while read -r item; do
-    echo "=========================="
-    echo "checking $item"
- 
-    case "$item" in
-    *"./package.json"* | *".github"* | *"functions/generated/client"*)
```
janpio, line 92: Interesting that we excluded a package.json from a generated Client.
I think we still have one of those, or even a new one.
Might this cause trouble?

scripts/update-ci.sh
millsp: Yes, I chose to exit earlier to avoid job overlaps which happens needlessly in the current setup sometimes

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-  v=$(bash .github/scripts/prisma-version.sh "$branch")
-  if [ -z "$v" ]
-  then
-        echo "Prisma version is empty: $v"
-        exit 0
-  fi
-  echo "$v (via Npm)"
-  
-  packages=$(find . -not -path "*/node_modules/*" -type f -name "package.json")
-  echo "$packages" | tr ' ' '\n' | while read -r item; do
```
janpio, line 87: Interesting how in every loop we seem to go through all projects, no matter if there is a "new" version or not - we check for each project. I think this ensures that outdated dependencies are updated as soon as possible, not only when a fully new version is released. Not sure that is worth or important to recreate.

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-  if [ -z "$(git status -s)" ]; then
-    echo "no changes"
-    end=$(date "+%s")
-    diff=$(echo "$end - $start" | bc)
-    remaining=$((interval - 1 - diff))
-    echo "took $diff seconds, sleeping for $remaining seconds"
-    sleep "$(no_negatives $remaining)"
-
-    continue
-  fi
```
janpio, line 137: This logic here is pretty fancy, as it ensured that we did sleep at least long enough to achieve the wait time intended no matter how long the individual loop ran. No idea why we did it this way, or if this is important.

.github/scripts/check-for-update.sh
millsp: I think this was the base package.json, pnpm handles that and the two others as well

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-
-  echo ""
-  echo "=========================="
-  echo "changes, upgrading..."
-  echo "$v" > .github/prisma-version.txt
-
-  git commit -am "chore(packages): bump prisma to $v"
```
janpio, line 144: This additional commit might be important - "bump" and "sync" were somehow separate

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-
-  set +e
-  git pull github "$branch" --rebase
-  code=$?
-  if [ $code -ne 0 ]; then
-    export webhook="$SLACK_WEBHOOK_URL_FAILING"
-    node .github/slack/notify.js "Prisma version $v :warning: Merge conflict at the end of check-for-update.sh script (via $branch)"
-    exit 0
-  fi
-  set -e
```
janpio, line 154: Can this not happen any more in the new setup?

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-    node .github/slack/notify.js "Prisma version $v released (via $branch)"
-  fi
-
-  end=$(date "+%s")
-  diff=$(echo "$end - $start" | bc)
-  remaining=$((interval - 1 - diff))
-  # upgrading usually takes longer than a few individual loop runs, so skip test runs which would have passed by now
-  skip=$((remaining / interval))
-  i=$((i - skip))
-  echo "took $diff seconds, skipping $skip x $interval second runs"
```
janpio, line 175: no idea how that worked honestly

.github/scripts/check-for-update.sh
```diff
@@ -1,178 +0,0 @@
...
-  code=$?
-  set -e
-  echo "pushed commit"
-
-  if [ $code -eq 0 ]; then
-    export webhook="$SLACK_WEBHOOK_URL"
-    node .github/slack/notify.js "Prisma version $v released (via $branch)"
-    export webhook="$SLACK_WEBHOOK_URL_FAILING"
-    node .github/slack/notify.js "Prisma version $v released (via $branch)"
-  fi
```
janpio, line 167: Replacement for this?

.github/scripts/check-for-update.sh
millsp: If I understood you correctly, this is why I also opted for replacing all the versions at once in all projects (discussed above), and then only install, so we can avoid this weird looping and the need to even check.

.github/scripts/check-for-update.sh
millsp: I don't think it matters anymore since we loop through the projects only if there is an update, we also don't diff anymore as a result of that since we know there was an update.

.github/scripts/check-for-update.sh
millsp: Yes, I think should restore that.

.github/scripts/check-for-update.sh
millsp: Will port that

.github/workflows/test.yaml
janpio: So it hard codes `package-lock.json` for any projects containing `azure` and `pnpm-lock.yaml` for all others?

platforms-serverless-vercel/vercel-node-builder/test.sh
janpio: Which means that the current test, that is running and passing is .... wrong?

platforms-serverless/firebase-functions/functions/package.json
janpio: new versions of firebase/gcp?

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
janpio: Probably because of the new custom output.

.github/workflows/upgrade-branch-to-version.yaml
janpio: I see no past runs, so was def not popular. Guess we can rebuild when/if we need it.

scripts/update-all.sh
janpio: That should get a comment - I don't think that will be obvious in a few days/weeks/months/years.

.github/scripts/check-for-update.sh
janpio: I think this still does more than your new script.
Yours only replaces _anything_ if there is a new version on Npm vs. what is recorded in the branch.
The old script you could commit a new project with an outdated version, and on next run of the scheduled CI it updated that one project to the _current_ version.

platforms-serverless-vercel/vercel-node-builder/test.sh
millsp: I think it means we have clear inconsistencies between package managers and we should sort that out. I had seen this before and wasn't sure, but now this looks like a clear confirmation that there is an issue. I'll create an issue.

platforms-serverless/firebase-functions/functions/package.json
millsp: Yes. Very hard to find in their docs.

platforms-serverless-vercel/vercel-node-builder/test.sh
millsp: (can this be marked as resolved?)

platforms/aws-graviton/code/index.test.js
millsp: (can this be marked as resolved?)

platforms-serverless/firebase-functions/functions/package.json
millsp: (can this be marked as resolved?)

platforms-serverless/netlify-github/package.json
millsp: (can this be marked as resolved?)

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
millsp: (can this be marked as resolved?)

platforms-serverless-vercel/vercel-node-builder/index.js
millsp: (can this be marked as resolved?)

scripts/update-all.sh
millsp: (can this be marked as resolved?)

scripts/update-all.sh
millsp: (can this be marked as resolved?)

.github/scripts/check-for-update.sh
millsp: (can this be marked as resolved?)

.github/scripts/check-for-update.sh
millsp: (can this be marked as resolved?)

.github/scripts/check-for-update.sh
millsp: (can this be marked as resolved?)

.github/scripts/check-for-update.sh
millsp: (can this be marked as resolved?)

.github/workflows/test.yaml
millsp: Turns out it is actually used for testing dependencies (eg. express)

.github/workflows/test.yaml
millsp: Yes

platforms/codesandbox/package.json
millsp: Turns out it is still needed to wake the sandbox up/start the dev server

scripts/update-all.sh
millsp: Added a comment

scripts/update-all.sh
millsp: Added a comment

platforms-serverless-vercel/vercel-node-builder/test.sh
millsp: Created an issue https://github.com/prisma/client-planning/issues/315

platforms/m1-macstadium/package.json
millsp: Currently pnpm will not create a lockfile if there are no dependencies, so that is a small quirk that happens in 2-3 projects. We use the lockfile to decide which package manager to run, that is convenient for most projects, so I think it's worth keeping this (even if technically not needed).

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
millsp: Chosen solution: test both npm and pnpm

platforms-serverless/serverless-framework-lambda/index.ts
millsp: https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1163418117

platforms-serverless/serverless-framework-lambda/serverless.yml
millsp: https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1163418117

platforms-serverless/serverless-framework-lambda/serverless.yml
millsp: https://github.com/prisma/ecosystem-tests/pull/3463#discussion_r1163418117

platforms-serverless/serverless-framework-lambda/prisma/schema.prisma
millsp: Ok, lambda permission didn't allow me to do that easily, so testing via npm only now

millsp:
To open all the resolved comments, load all of them and then run 
```js
const buttons = document.getElementsByClassName('Details-content--closed'); let i = 0; while(buttons.length != 0) buttons[i++].click();
```

millsp:
`serverless-framework-lambda` is very flaky because the two async jobs (binary, library) overwrite each other. I don't have access to aws/serverless.com so I could not create an extra service/function for binary/library.

PR #3464, opened 2023-04-05
feat(docker): add debian-bullseye-amd64-openssl-1.1.x-tls

aqrln:
Fails with library engine with the following error before https://github.com/prisma/prisma/pull/18426:

```
Error opening a TLS connection: error:0308010C:digital envelope routines:func(128):unsupported:../deps/openssl/openssl/crypto/evp/evp_fetch.c:373:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
```

(It doesn't segfault because it takes a different code path: here I used the Google Cloud PostgreSQL database we already had in the tests, which uses client certificates, so it fails a bit earlier with a different error, but the root cause is essentially the same as https://github.com/prisma/prisma/issues/10649. I didn't find any other database in our test infra that would be affected to reproduce precisely that issue and I wanted to avoid adding a new database to manage.)

Doesn't fail in any way after https://github.com/prisma/prisma/pull/18426.


millsp:
Should we get this merged?

aqrln:
yes, it's a regression test for an issue we previously fixed so would be nice to merge

Jolg42 (approved):

PR #3465, opened 2023-04-10
feat: Json protocol (dev)

janpio:


janpio:
Not viable right now because pnpm is still broken in `dev` :/

PR #3486, opened 2023-05-05
feat(docker): restore tests for debian stretch

aqrln:
- Revert "feat(docker): remove debian stretch tests, as it reached EOL (#3485)"
- Change the apt sources to archive since Stretch is EOL


Jolg42 (approved):
Welcome back, logs look good

PR #3500, opened 2023-05-08
test(engines): add engine-native-deps

aqrln:
Check that the native dependencies of Prisma Engines on Linux match what we expect.

Closes: https://github.com/prisma/client-planning/issues/370


.github/scripts/test-project.sh
```diff
@@ -119,8 +119,12 @@ elif [ $code -eq 0 ]; then
 
   # confirm existence of correct engine
   echo "-------------- Checking Engines ----------------"
-  bash ../../.github/scripts/check-engines-client.sh $dir $project
-  bash ../../.github/scripts/check-engines-cli.sh $dir $project
+  if [ -z "${SKIP_ENGINE_CHECK+x}" ]; then
```
Jolg42, line 122: What does `+x` do here?

aqrln: this is just to avoid the unbound variable error because of `set -u`

Jolg42 (approved):
Looks good!

Latest logs: https://github.com/prisma/ecosystem-tests/actions/runs/4918642778/jobs/8785267147?pr=3500#step:7:1

jkomyno (approved):

PR #3510, opened 2023-05-11
fix(notify): add retries for slack notifications

millsp:
Because tests are finishing faster since the refactor to `pnpm`, we are now hitting rate limits for pushing slack notifications. This PR fixes that by adding a retry with exponential backoff.


Jolg42 (approved):
Didn't expect that one 😄

janpio:
We could/should probably also remove posting of successful runs/workflows/jobs like #feed-ecosystem-tests-jobs

Jolg42:
@janpio Indeed, nobody is looking at the successful notifications since a long time now 😄

millsp:
Removed 👨‍⚖️

PR #3569, opened 2023-05-31
chore: split cosmosdb failing test into weekly workflow

### LINKED ISSUE #3351: Split out expected failing tests from optional tests so it is clearer that they are expected to fail

Jolg42:
Closes https://github.com/prisma/ecosystem-tests/issues/3351

### DIFF
--- a/.github/scripts/check-folders.js
+++ b/.github/scripts/check-folders.js
@@ -3,14 +3,12 @@ const yaml = require('yaml')
 
 const fs = require('fs')
 const path = require('path')
-const { exit } = require('process')
 
 /**
  * This script checks if there are folders (= tests) that are not "run" via test.yaml and test.yaml
- *  It gets all relevant folders, then all the jobs from both workflows, and extract those from the list of folder and checks if nothing is left 
+ *  It gets all relevant folders, then all the jobs from both workflows, and extract those from the list of folder and checks if nothing is left
  */
 async function main() {
-  
   // Files and folders to skip during checking
   const ignoreFiles = [
     'package.json', // package.json at root
@@ -23,8 +21,9 @@ async function main() {
     'packagers/yarn3-workspaces-pnp/packages/sub-project-2', // We don't want to include the workspace folders in the matrix
     'platforms/aws-graviton/code', // aws-graviton doesn't have package.json at root but is included
     'platforms/m1-macstadium/code', // m1-macstadium doesn't have package.json at root but is included
+    'databases/mongodb-azure-cosmosdb', // expected to fail in failing-weekly.yaml
   ]
-  
+
   // Jobs in the workflow files that are not relevant and can be skipped
   const jobsToIgnore = [
     'start-time', // Not a test but a job that fills an env var with the job start time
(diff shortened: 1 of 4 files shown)


.github/workflows/failing-weekly.yaml
```diff
@@ -0,0 +1,60 @@
+name: failing-weekly
+
+on:
+  workflow_dispatch:
+  schedule:
+    # at 3am weekly on Sunday
+    - cron: '0 3 * * 0'
+
```
miguelff, line 8: Would it be ~sensible~ convenient to commit a change to run on PR push so we can see the test fail, and then revert that commit? That would provide feedback in the PR of the expected outcome.

.github/scripts/check-folders.js
```diff
@@ -23,8 +21,9 @@ async function main() {
     'packagers/yarn3-workspaces-pnp/packages/sub-project-2', // We don't want to include the workspace folders in the matrix
     'platforms/aws-graviton/code', // aws-graviton doesn't have package.json at root but is included
     'platforms/m1-macstadium/code', // m1-macstadium doesn't have package.json at root but is included
+    'databases/mongodb-azure-cosmosdb', // expected to fail in failing-weekly.yaml
```
janpio: this is an actual change in this file vs only whitespace/formatting changes

miguelff (approved):

.github/workflows/failing-weekly.yaml
janpio: So that it runs on any `dev` version of `prisma/prisma`?

That is currently how it is implemented in our main workflow - but the "failing" tests we _know_ do not succeed because of a - in this case - limitation of the database. So this test is tracking if an update of the database surprisingly fixes that limitation.

.github/workflows/failing-weekly.yaml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  databases:
+    strategy:
+      fail-fast: false
+      matrix:
+        database:
+          # It fails, see https://github.com/prisma/prisma/issues/15083
+          - mongodb-azure-cosmosdb
+        # Since it fails, we can test only library
```
janpio, line 26: Why?

janpio: Ah, not in "we are only able to test library", but "it is enough to test one of both"?

.github/scripts/check-folders.js
janpio:
```suggestion
    'databases/mongodb-azure-cosmosdb', // expected to fail, so moved to failing-weekly.yaml
```

janpio:
```suggestion
    'databases/mongodb-azure-cosmosdb', // expected to fail, so moved to failing-weekly.yaml
```

janpio: Alternatively to listing it here, we could also add reading `failing-weekly` instead, right?

janpio (approved):

.github/workflows/failing-weekly.yaml
Jolg42: It fails the same, the issue is not about the engine here.

.github/scripts/check-folders.js
Jolg42: Probably, but I suspect it's going to be the only one anyway

.github/workflows/failing-weekly.yaml
Jolg42: We already know it fails and why and want to move it away but keep it around to know if maybe one day it succeeds.

PR #3692, opened 2023-07-12
test(node): 16.13 using @pnpm/exe with Node packaged

Jolg42:
https://prisma-company.slack.com/archives/CRGDU19K6/p1689163780627219

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -635,7 +635,7 @@ jobs:
       fail-fast: false
       matrix:
         node:
-           - 16.13 # Prisma's minimum is 16.13 but pnpm 8 requires Node.js 16.14, so we test it separately, see `node16dot13` 
+          - 16.13 # Prisma's minimum is 16.13 but pnpm 8 requires Node.js 16.14, so we test it separately, see `node16dot13` 
           - 16
           - 17
           - 18
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -573,6 +573,55 @@ jobs:
         if: failure()
         run: bash .github/slack/notify-failure.sh generic basic ${{ matrix.os }}
 
+  # pnpm requires Node.js v 16.14 or higher
+  # so as a workaround we test this version with npm
+  node16dot13:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -581,7 +635,6 @@ jobs:
       fail-fast: false
       matrix:
         node:
-          - 16.14 # Prisma's minimum is 16.13 but pnpm 8 minimum is 16.14
```
janpio, line 584: Let's keep this line around but commented so we know how we _should_ be testing this. Otherwise this will get lost sooner or later and we will not explicitly be tested the minimum version any more - and that is important.

.github/workflows/test.yaml
```diff
@@ -581,7 +635,6 @@ jobs:
       fail-fast: false
       matrix:
         node:
```
Jolg42, line 637:
```suggestion
        node:
           - 16.13 # Prisma's minimum is 16.13 but pnpm 8 requires Node.js 16.14, so we test it separately, see `node16dot13` 
```

.github/workflows/test.yaml
```diff
@@ -581,7 +635,7 @@ jobs:
...
-          - 16.14 # Prisma's minimum is 16.13 but pnpm 8 minimum is 16.14 
+          - 16.13 # Prisma's minimum is 16.13 but pnpm 8 requires Node.js 16.14, so we test it separately, see `node16dot13` 
           - 16
```
```suggestion
          # Prisma's minimum is 16.13 but pnpm 8 requires Node.js 16.14, so we test it separately, see `node16dot13` - which is why the next line is commented out
          # - 16.13 # minimal node version via https://github.com/prisma/prisma/blob/main/packages/client/package.json and minimal minor+patch version of node via https://www.prisma.io/docs/reference/system-requirements
          - 16 # latest minor+patch on GH Actions
```

Jolg42: "and minimal minor+patch version of node 14 via ", 🤔 why 14?

janpio: because copy/paste

janpio: updated

PR #3721, opened 2023-07-19
test(codesandbox): remove --preview-feature from db push

Jolg42:
It was easy...!

Note that we don't have a codesanbox account since we use the public API


platforms/codesandbox/deploy.ts
```diff
@@ -36,7 +36,7 @@ async function fetchWithPuppeteer(sandboxId: string) {
   const page = await browser.newPage()
   await page.setDefaultNavigationTimeout(0)
   await page.goto(`https://codesandbox.io/p/sandbox/${sandboxId}`)
-  await page.waitForTimeout(20_000)
+  await sleep(20)
```
Jolg42, line 39: waitForTimeout is deprecated

millsp (approved):

PR #3820, opened 2023-08-18
fix: adapt tests for pdp client changes

millsp:
closes https://github.com/prisma/team-orm/issues/310

### DIFF
--- a/dataproxy/vercel-cli-serverless-functions/run.sh
+++ b/dataproxy/vercel-cli-serverless-functions/run.sh
@@ -35,8 +35,8 @@ else
   exit 1
 fi
 
-# Check the Vercel Build Logs for "| dataproxy" in generate
-if echo "${OUTPUT}" | grep -q '| dataproxy'; then
+# Check the Vercel Build Logs for "engine=none" in generate
+if echo "${OUTPUT}" | grep -q 'engine=none'; then
   echo 'Data Proxy was successfully enabled'
 else
   echo "Data Proxy was NOT ENABLED"
--- a/platforms-serverless/netlify-cli/test.sh
+++ b/platforms-serverless/netlify-cli/test.sh
@@ -3,9 +3,9 @@
 set -eux
 
 if [ "$PRISMA_CLIENT_ENGINE_TYPE" == "binary" ]; then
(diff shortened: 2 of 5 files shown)


platforms-serverless/netlify-cli/test.sh
```diff
@@ -3,9 +3,9 @@
 set -eux
 
 if [ "$PRISMA_CLIENT_ENGINE_TYPE" == "binary" ]; then
-  BINARY_STRING=',"files":["index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","runtime","schema.prisma"]'
+  BINARY_STRING=',"files":["edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","package.json","query-engine-rhel-openssl-1.0.x","runtime","schema.prisma"]'
```
janpio, line 6: These are now always included in the Client?

(Answer as comment, not emoji reaction please so I get notification. Thanks.)

dataproxy/vercel-cli-serverless-functions/run.sh
```diff
@@ -35,8 +35,8 @@ else
   exit 1
 fi
 
-# Check the Vercel Build Logs for "| dataproxy" in generate
-if echo "${OUTPUT}" | grep -q '| dataproxy'; then
+# Check the Vercel Build Logs for "engine=none" in generate
+if echo "${OUTPUT}" | grep -q 'engine=none'; then
```
janpio, line 39: New output in `generate`?

(Answer as comment, not emoji reaction please so I get notification. Thanks.)

platforms-serverless/netlify-cli/test.sh
millsp: Yes, that's the side-effect of the removal of `--data-proxy` and also because it is a custom output.

dataproxy/vercel-cli-serverless-functions/run.sh
millsp: Indeed.

platforms-serverless/netlify-cli/test.sh
janpio: Surprising that this did not affect more files/expectations.

millsp: Non-custom output setups already had an edge client that just throws an error so the file already existed for these, but that did not exist for custom output projects.

PR #3875, opened 2023-08-30
test(client): driver adapters on vercel via next.js (neon, planetscale, pg)

millsp:


driver-adapters/neon-vercel-nextjs/prisma/schema.prisma
```diff
@@ -0,0 +1,15 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DRIVER_ADAPTERS_NEON_VERCEL_NEXTJS_DATABASE_URL")
```
janpio: It might be good idea to set this to an invalid value to make sure it is not being used under the hood instead of the _actual_ driver with the defined-in-code value.

janpio (approved):

janpio (approved):

Jolg42:
Note: from what I understand, this should fix the unrelated failures https://github.com/prisma/ecosystem-tests/pull/3880

.github/workflows/test.yaml
```diff
@@ -1122,6 +1122,75 @@ jobs:
         if: failure()
         run: bash .github/slack/notify-failure.sh ${{ github.job }} ${{ matrix.platform }}
 
+  driver-adapters:
+    needs: [detect_jobs_to_run]
+    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['driver-adapters'] == true }}
+
+    timeout-minutes: 60 # can take longer if platforms are down, so better protect
```
```suggestion
    timeout-minutes: 60 # can take longer if platforms are down, so better protect
```
Note: 60 minutes sounds like a lot to me for these tests 🤔

.github/workflows/test.yaml
```diff
@@ -1122,6 +1122,75 @@ jobs:
...
+        combo:
+          - neon-vercel-nextjs
+          - planetscale-vercel-nextjs
+          - postgresql-vercel-nextjs
+        clientEngine: ['library']
+        os: [ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
+    concurrency: driver-adapters-${{ matrix.combo }}-${{ matrix.clientEngine }}
+    env:
+      NPM_CONFIG_LOGLEVEL: error
```
Jolg42: I'm not sure why we had this set to `error`, `warn` would be better?

Jolg42 (approved):

.github/workflows/test.yaml
millsp: I removed it

PR #3880, opened 2023-08-31
fix: broken tests and do some house keeping

millsp:
* Fixes broken tests
* Upgrades redwood
* Removes unused test


.npmrc
```diff
@@ -3,3 +3,7 @@ shared-workspace-lockfile=false
 
 # without this, installing one project installs all
 recursive-install=false
+
+# so client postinstall hook is not cached for all
+# https://github.com/pnpm/pnpm/releases/tag/v8.7.1
+side-effects-cache=false
```
millsp, line 9: Does not actually fix the issue because we have dependency caching, but works if there's no caching.

docker/almalinux-8-amd64-openssl-1.1.x/run.sh
```diff
@@ -30,3 +30,4 @@ docker run -p 3000:3000 \
 sleep 15
 
 pnpm install
+pnpm prisma generate
```
millsp, line 33: I rather add it here instead of having a custom project exclusion.

platforms-serverless/lambda/test.ts
```diff
@@ -22,9 +22,9 @@ async function main() {
   const actual = JSON.stringify(original)
   console.log('actual', actual)
   // TODO Update to only expect on engine file after zip script was updated
-  let files = `,"files":["deno","edge.d.ts","edge.js","index-browser.js","index.d.ts","index.js","libquery_engine-debian-openssl-1.1.x.so.node","libquery_engine-rhel-openssl-1.0.x.so.node","package.json","schema.prisma"]`
```
millsp, line 25: This was incorrect, not sure why. `deno` is only created when the client creates the stub files, but not when there is a valid schema that can be generated from.

platforms-serverless/serverless-framework-lambda/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "outDir": ".",
     "strict": true,
     "lib": ["esnext", "dom"],
-    "esModuleInterop": true
```
millsp, line 8: To suppress a type issue in `lodash`.

platforms-serverless/serverless-framework-lambda/tsconfig.json
```diff
@@ -5,6 +5,7 @@
...
+    "esModuleInterop": true,
+    "skipLibCheck": true,
```
millsp, line 10: To suppress a type issue in `lodash` itself.

millsp:
Trying to fix Redwood, among other things, the tests are not passing because of other [reasons](https://prisma-company.slack.com/archives/CL0AJEVSA/p1693528780922799) (both fail on Redwood 3 (previous), and Redwood 6 (this PR)).

platforms-serverless/serverless-framework-lambda/tsconfig.json
Jolg42: Make this a comment then, it's ok to have a comment in tsconfig
```suggestion
    // workaround for a type issue in lodash itself
    "skipLibCheck": true,
```

platforms-serverless-vercel/vercel-with-redwood/api/jsconfig.json
```diff
@@ -1,11 +1,11 @@
 {
   "compilerOptions": {
     "noEmit": true,
-    "allowJs": true,
     "esModuleInterop": true,
     "target": "esnext",
     "module": "esnext",
     "moduleResolution": "node",
+    "skipLibCheck": false,
```
Jolg42, line 8: Why is it needed for Redwood?

platforms-serverless-vercel/vercel-with-redwood/package.json
```diff
@@ -22,8 +21,8 @@
     "root": true
   },
   "engines": {
-    "node": "16.x",
-    "yarn": ">=1.15 <2"
+    "node": "=16.x",
```
Jolg42: Why not
```suggestion
    "node": ">=16.x",
```

Jolg42 (approved):
Looks good to me! Interesting to see that something is broken with Redwood/Vercel

binaries/pkg/run.sh
```diff
@@ -3,6 +3,7 @@
 set -eu
 
 pnpm install
+pnpm prisma generate
```
janpio, line 6: Why is this needed explicitly?

docker/almalinux-8-amd64-openssl-1.1.x/run.sh
janpio: Why is all this needed? This should happen automatically, under the hood, right?

millsp: Because our dependencies are cached to speed up the runs, we end up in a situation where the `postinstall` of Prisma Client is not always triggered, similar to what we had with Vercel.

platforms-serverless-vercel/vercel-with-redwood/api/jsconfig.json
millsp: It was added by Redwood, not sure. Maybe for performance?

binaries/pkg/run.sh
millsp: Because our dependencies are cached to speed up the runs, we end up in a situation where the postinstall of Prisma Client is not always triggered, similar to what we had with Vercel.

platforms-serverless-vercel/vercel-with-redwood/api/jsconfig.json
Jolg42: I see, and less problems from users too maybe 😄

PR #3881, opened 2023-08-31
repro https://github.com/prisma/prisma/issues/20905

janpio:

PR #3882, opened 2023-08-31
repro https://github.com/prisma/prisma/issues/20905 #2

janpio:
Same as https://github.com/prisma/ecosystem-tests/pull/3881 but for `latest` (5.2.0)


janpio:
Ok, this is still on 20.x - so might not be exact same situation as from user.

PR #3898, opened 2023-09-04
fix(driver-adapters): prevent version bumping

millsp:
Prevents the default bumping of Prisma and Prisma Client in the `driver-adapters` sub-project so we can stick with the integration version for now.

### DIFF
--- a/packagers/yarn-workspaces/yarn.lock
+++ b/packagers/yarn-workspaces/yarn.lock
@@ -27,22 +27,22 @@
     "@jridgewell/resolve-uri" "^3.0.3"
     "@jridgewell/sourcemap-codec" "^1.4.10"
 
-"@prisma/client@5.3.0-dev.36":
-  version "5.3.0-dev.36"
-  resolved "https://registry.yarnpkg.com/@prisma/client/-/client-5.3.0-dev.36.tgz#a28e12787a5f4f7c9e4154c24d94e4cf26df6aee"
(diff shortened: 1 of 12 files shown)


packagers/yarn-workspaces/yarn.lock
```diff
@@ -99,12 +99,12 @@ make-error@^1.1.1:
...
-prisma@5.3.0-dev.36:
-  version "5.3.0-dev.36"
-  resolved "https://registry.yarnpkg.com/prisma/-/prisma-5.3.0-dev.36.tgz#84dec32fce7be4bd5c12263f6ad1c277a6ac4bf1"
-  integrity sha512-MEgndZlSJxCeZkkwDwJhikwens3i557e5p2QyGsTkhwVmX62A4lozqO3Wkq09xFKhcI+MGx7brFx7CNEe4+Ijg==
+prisma@5.3.0-dev.38:
+  version "5.3.0-dev.38"
+  resolved "https://registry.yarnpkg.com/prisma/-/prisma-5.3.0-dev.38.tgz#af352fda2f5d487077bd1631dee1fa21b4de889e"
+  integrity sha512-Rmlhqd4+ieqDQyWOSqz+U3s6I0QhwB82enEEFgZJIWIHu+HFk4pPDyxm70kWiNetpQLPB7rHt7jB54i9rWPPQQ==
   dependencies:
-    "@prisma/engines" "5.3.0-dev.36"
```
millsp, line 107: Flakiness with yarn that I just noticed, this is because we run it in parallel to update lockfiles but it does not work very well. Will fix in a separate commit.

PR #3918, opened 2023-09-07
test(dataproxy, edge): remove /edge imports

millsp:


millsp:
You can still review, I just thought now that we should test both with `@prisma/client/edge` and `@prisma/client`, so I think I need to add that.

dataproxy/cloudflare-workers-module/package.json
```diff
@@ -5,7 +5,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "@prisma/client": "5.3.0-dev.52",
+    "@prisma/client": "5.3.0-integration-feat-auto-edge-import.4",
```
Jolg42, line 8: TODO: before merging, switch back to dev version

janpio:
Replaced by https://github.com/prisma/ecosystem-tests/pull/4192

PR #3928, opened 2023-09-11
tests: adapt for new Accelerate itx timeout at 15s

janpio:
context see https://prisma-company.slack.com/archives/C04F60M3G06/p1694342753774529

PR #3929, opened 2023-09-11
fix: DP tests, MongoDB and order

janpio:
follow up to https://github.com/prisma/ecosystem-tests/pull/3928


.github/workflows/test.yaml
```diff
@@ -1298,33 +1300,34 @@ jobs:
           PRISMA_GENERATE_FLAG: '--accelerate'
           DATAPROXY_COMMON_URL: ${{ secrets.DP2_DATAPROXY_COMMON_URL }}
           ITX_PDP_MONGODB_DATABASE_URL: ${{ secrets.ITX_PDP_MONGODB_DATABASE_URL }}
-          ITX_PDP_MONGODB: ${{ secrets.ITX_PDP_MONGODB }} # MongoDB is not yet supported in DP2
+          ITX_PDP_MONGODB: ${{ secrets.ITX_PDP_MONGODB }}
```
janpio: This should be changed to an Accelerate connection string

janpio:
`DP2_ITX_PDP_MONGODB` is currently setup as an org secret - we need to move that (value see 1pwd)

PR #3932, opened 2023-09-12
chore(driver-adapters): update to latest dx

millsp:
https://www.notion.so/prismaio/How-to-Prisma-Client-Driver-Adapters-d13f6c0563824206befeb78a6bbb67d0

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -38,7 +38,7 @@ else
   default_version="$(cat .github/prisma-version.txt)"
   cli_version_dev="$(node -e "console.log(require('./$dir/$project/package.json').devDependencies.prisma ?? '')")"
   cli_version_dep="$(node -e "console.log(require('./$dir/$project/package.json').dependencies.prisma ?? '')")"
-  version=$(node -e "console.log('$cli_version_dev' || '$cli_version_dep' || '$default_version')")
+  version=e$(node -e "console.log('$cli_version_dev' || '$cli_version_dep' || '$default_version')")"
 
   schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
   if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -36,8 +36,8 @@ then
   # if a project needs to be skipped for any reason, replace `foobar` with its folder name or add additional conditions
 else
   default_version="$(cat .github/prisma-version.txt)"
-  cli_version_dev="$(node -e "console.log(require('./$dir/$project/package.json').devDependencies.prisma ?? '')")"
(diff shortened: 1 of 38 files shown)


.github/scripts/test-project.sh
```diff
@@ -28,6 +28,13 @@ echo ""
 echo "-----------------------------"
 echo "running $dir/$project"
 
+echo "cd $dir/$project"
+cd "$dir/$project"
+
+PM=$(node -e "require('@antfu/ni').detect({ autoInstall: false, cwd, programmatic: true }).then(console.log)")
+INSTALL_CMD=$(node -e "require('@antfu/ni').getCommand('$PM', 'install', []).then(console.log)")
+
+sh -c "$INSTALL_CMD"
```
millsp, line 37: Prepares the project before had, so it can use the installed cli version of the project.

.github/scripts/test-project.sh
```diff
@@ -40,14 +47,11 @@ else
     echo ""
     echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
     echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
-    npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
+    sh -c "$(node -e "console.log(require('@antfu/ni').getCommand('$PM', 'execute', ['prisma', 'db', 'push', '--accept-data-loss', '--skip-generate', '--schema=$schema_path']))")"
```
millsp, line 50: Executes the migration with the right package managers.

.github/scripts/test-project.sh
```diff
@@ -35,12 +35,17 @@ then
...
+  version="$(node -e "console.log('$cli_version_dev' || '$cli_version_dep' || '$default_version')")"
+
   schema_path=$(find $dir/$project -name "schema.prisma" ! -path "*/node_modules/*" | head -n 1)
   if grep -q "env(\"DATABASE_URL\")" "$schema_path"; then
     echo ""
     echo "found 'schema.prisma' with 'env(\"DATABASE_URL\")': $schema_path"
-    echo "npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path"
-    npx prisma db push --accept-data-loss --skip-generate --schema=$schema_path
+    echo "pnpm dlx prisma@$version db push --accept-data-loss --skip-generate --schema=$schema_path"
+    pnpm dlx prisma@$version db push --accept-data-loss --skip-generate --schema=$schema_path
```
millsp: Small addition so that the integration version is used in the project, instead of the globally installed `prisma`. This is better for consistency when working with with integration versions and ensures (best effort) that we are using the integration end-to-end, even in automatic db creation.

millsp: This is actually needed, because `prisma@5.2.0` is not aware of the `driverAdapters` preview feature.

Jolg42 (approved):
Nice ✨

PR #3939, opened 2023-09-12
fix: DP tests, MongoDB and order

janpio:
replaces #3929

Note: New env vars like `DP2_ITX_PDP_MONGODB` are currently setup as an org secret - we need to move that (value see 1pwd)


janpio:
See https://github.com/prisma/ecosystem-tests/issues/3944

PR #4177, opened 2023-10-29
fix(databases/docker-pgbouncer): Make it actually test what it should be testing

janpio:


connection-poolers/docker-pgbouncer/prisma/schema.prisma
```diff
@@ -5,6 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
+  engineType = "binary"
```
janpio, line 8: Needed as behavior to test only works with binary.

PR #4193, opened 2023-11-01
feat: turso-embedded-replica

janpio:


driver-adapters/turso-embedded-replica-node-basic/index.js
```diff
@@ -0,0 +1,182 @@
+// @ts-check
+const { Prisma, PrismaClient } = require('@prisma/client')
+const { createClient } = require('@libsql/client')
+const { PrismaLibSQL } = require('@prisma/adapter-libsql')
+
+const connectionString = process.env.DRIVER_ADAPTERS_TURSO_NODE_BASIC_DATABASE_URL
```
janpio, line 6: This should probably use a different database.

PR #4303, opened 2023-11-25
feat: `driver-adapters-wasm` folder + `pg-node-basic-wasm(-edge-import)` project

janpio:
Tests that uses the wasm engine (with both possible imports) locally via Node.
Also splits all the Wasm tests into their own job so it is easier to keep them apart.


.github/scripts/check-engines-client.sh
```diff
@@ -1,46 +1,41 @@
...
-if [ -z "$PRISMA_CLIENT_ENGINE_TYPE" ]; then
-  echo "Using default client engine: $DEFAULT_CLIENT_ENGINE_TYPE"
-  CLIENT_ENGINE_TYPE=$DEFAULT_CLIENT_ENGINE_TYPE
-else
-  echo "Using env(PRISMA_CLIENT_ENGINE_TYPE): $PRISMA_CLIENT_ENGINE_TYPE"
-  CLIENT_ENGINE_TYPE=$PRISMA_CLIENT_ENGINE_TYPE
-fi
-
-# These are skipping for a variety of reasons like:
+# These are beinf skipped for a variety of reasons like:
```
janpio:
```suggestion
# These are being skipped for a variety of reasons like:
```

.github/scripts/test-project.sh
```diff
@@ -122,15 +122,18 @@ elif [ $code -eq 0 ]; then
   echo ""
   echo "-----------------------------"
 
+  echo ""
+  echo ""
+
   # confirm existence of correct engine
-  echo "-------------- Checking Engines ----------------"
+  echo "-------------- Checking Engines -----------------------------------"
```
janpio:
```suggestion
  echo "-------------- Checking Engines -------------------------------"
```

.github/scripts/test-project.sh
```diff
@@ -122,15 +122,18 @@ elif [ $code -eq 0 ]; then
...
   if [ -z "${SKIP_ENGINE_CHECK+x}" ]; then
     bash ../../.github/scripts/check-engines-client.sh $dir $project
     bash ../../.github/scripts/check-engines-cli.sh $dir $project
   else
     echo "SKIP_ENGINE_CHECK=$SKIP_ENGINE_CHECK, skipping"
   fi
-  echo "------------------------------------------------"
+  echo "-------------------------------------------------------------------"
```
janpio:
```suggestion
  echo "---------------------------------------------------------------"
```

.github/scripts/check-engines-client.sh
janpio, line 1: Opportunistic improvements to make the output easier to understand.

.github/scripts/test-project.sh
janpio, line 1: Same here.

.github/workflows/test.yaml
```diff
@@ -920,7 +923,7 @@ jobs:
         clientEngine: ['library', 'binary']
         os: [ubuntu-20.04] #, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    concurrency: platforms-${{ matrix.platform }}-${{ matrix.clientEngine }}
+    concurrency: ${{ github.job }}-${{ matrix.platform }}-${{ matrix.clientEngine }}
```
janpio, line 944: This was a potential source for making mistakes. When `concurrency` is just job specific, let's use the job name automatically instead of string.

driver-adapters-wasm/pg-node-basic-wasm-edge-import/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "pg-node-basic-wasm-edge-import",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest index.test.mjs"
```
janpio: This (and some file renames, change from require to import) was needed to be able to run the tests, modelled after the other `esm` project under `driver-adapters`.

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
```diff
@@ -0,0 +1,160 @@
+// @ts-check
+import { Prisma, PrismaClient } from '@prisma/client'
+import { PrismaPg } from '@prisma/adapter-pg'
+import pkg from 'pg'
+const { Pool } = pkg
```
janpio: Can this be done cleaner?

.github/scripts/check-engines-client.sh
```diff
@@ -1,46 +1,41 @@
...
-  prisma-dbml-generator                   # No generated Client, so only Client stub with no engine included
-  prisma-json-schema-generator            # No generated Client, so only Client stub with no engine included
-  engine-types                            #
-  pkg                                     # No generated Client, so only Client stub with no engine included
-  aws-graviton                            # No local project at all (everything happens on server), so no `prisma` or `node_modules
-  firebase-functions                      # No local project at expected location (but in `functions` subfolder)
-  studio                                  # TODO: No generated Client in `node_modules/.prisma/client/`
-  netlify-cli                             # Client is generated into `../functions/generated/client` via use of `output`
-  jest-with-multiple-generators           # No generated Client locally in default path, both Clients have custom `output`
+  engine-types                                # No PRISMA_CLIENT_ENGINE_TYPE
```
millsp, line 15: Random idea: it would be nice to have these as a config into the project `package.json` maybe

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
```diff
@@ -0,0 +1,160 @@
...
+  })
+
+  const regResult = await getResult(prisma).catch((error) => ({ error: error.message }))
+  const itxResult = await prisma.$transaction(getResult).catch((error) => ({ error: error.message }))
+  const result = { itxResult, regResult }
+
+  return result
+}
+
+//console.log(await handler())
```
Jolg42:
```suggestion
console.log(await handler())
```

Is there a reason to comment it?

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
Jolg42: Yes!
```suggestion
import { Pool } from 'pg'
```

.github/workflows/test.yaml
```diff
@@ -1282,6 +1275,86 @@ jobs:
...
+          - neon-http-cf-basic
+          - neon-cf-basic
+          - planetscale-cf-basic
+          - turso-cf-basic
+        clientEngine: ['wasm']
+        os: [ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
+    concurrency: ${{ github.job }}-${{ matrix.combo }}-${{ matrix.clientEngine }}
+    env:
+      NODE_ENV: development
```
Jolg42, line 1321: Is `NODE_ENV` really needed here?

.github/scripts/check-engines-client.sh
```diff
@@ -127,11 +132,15 @@ else
...
   echo "✔ Data Proxy has no Query Engine" # TODO: actually check that there isn't one
 elif [ -f "$qe_location" ]; then
-  echo "✔ Correct Query Engine exists"
+  echo "✔ Correct Query Engine exists at ${qe_location}"
 else
   echo "❌ Could not find Query Engine in ${qe_location} when using ${os_name}"
   exit 1
 fi
+
+echo "$ ls ${qe_location}"
```
Jolg42:
```suggestion
echo "$ ls $GENERATED_CLIENT"
```

Jolg42 (approved):
Nice ✨ 

See small comments

Jolg42:
Advice: 
- disable https://github.com/prisma/ecosystem-tests/actions/workflows/check-for-update.yaml before updating versions to resolve the merge conflict
- merge
- enable it again

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
janpio: I think that is how I started, and then worked through the error messages until I ended up where I did :D I wonder what went wrong...

.github/scripts/check-engines-client.sh
janpio: I also thought about this, but then also decided tht it's not worth it at all as we almost never need to add things here. And when we need, the config is close to the failing script.

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
janpio: I think it ran automatically during tests somehow.

.github/workflows/test.yaml
janpio: No idea, that is just copy-pasta. @millsp?

janpio:
Actually, I finally figured out how to "resolve" most the version conflicts: 
Running `pnpm update-all 5.7.0-dev.54` with the version that is currently in `dev` fixes the problem for all the files outside of the PR.

Jolg42:
@janpio yes, but it's not guaranteed that you will be faster than our CI. What happens often is that while doing this another dev version shows up 😄

janpio:
True.
And now GH does not detect/show my push :/ 

Edit: This explains it: https://www.githubstatus.com/incidents/66vhjmd266r9

.github/workflows/test.yaml
```diff
@@ -1231,49 +1225,125 @@ jobs:
...
+      NODE_MODULES_CACHE: false
+      NODE_VERBOSE: true
+      CLOUDFLARE_ACCOUNT_ID: ${{ vars.CLOUDFLARE_ACCOUNT_ID }}
+      CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      DEBUG: 'prisma:js-connector*'
+      PRISMA_DISABLE_QUAINT_EXECUTORS: true
```
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -1231,49 +1225,125 @@ jobs:
...
+      DRIVER_ADAPTERS_TURSO_CF_BASIC_DATABASE_URL: ${{ vars.DRIVER_ADAPTERS_TURSO_CF_BASIC_DATABASE_URL }}
+      DRIVER_ADAPTERS_TURSO_CF_BASIC_TOKEN: ${{ secrets.DRIVER_ADAPTERS_TURSO_CF_BASIC_TOKEN }}
       DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_DATABASE_URL: ${{ vars.DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_DATABASE_URL }}
       DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_TOKEN: ${{ secrets.DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_TOKEN }}
       DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_PROJECT_ID: ${{ vars.DRIVER_ADAPTERS_TURSO_VERCEL_NEXTJS_EDGEMW_PROJECT_ID }}
-      # settings
-      DEBUG: 'prisma:js-connector*'
-      PRISMA_DISABLE_QUAINT_EXECUTORS: true
-
+
```
janpio, line 1349:
```suggestion
      # settings
      DEBUG: 'prisma:js-connector*'
      PRISMA_DISABLE_QUAINT_EXECUTORS: true
    
```

.github/workflows/test.yaml
millsp: It's probably not used indeed

driver-adapters-wasm/pg-node-basic-wasm-edge-import/index.mjs
millsp: It's not needed because the test calls handler

.github/workflows/test.yaml
millsp: We should remove all of them via a separate PR

millsp:
I noticed, the `driver-adapters-wasm` tests were [skipped](https://binaries.prisma.sh/all_commits/01aad9b63c8d574cc270d2b09461e920d19986e6/debian-openssl-1.1.x/libquery_engine.so.node.sha256) somehow, I'll investigate.

Update: It seems like there was a change in the detection logic, this is now needed instead
```
contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), ...)
```

millsp:
I renamed renamed `pg-node-basic-wasm` to `pg-node-basic` and renamed `pg-node-basic-wasm-edge-import` to `pg-node-basic-esm`. `wasm` was redundant in this new `driver-adapters-wasm` folder, and `edge-import` looked like it meant `esm`.

janpio:
Change in `if` logic was https://github.com/prisma/ecosystem-tests/pull/4307

PR #4307, opened 2023-11-26
ci: Simplify `detect_jobs_to_run` output (and related `if`)

janpio:
Checking the object was always a bit awkward, so I changed it to just write an array of the needed jobs instead of all jobs with true/false. I think that simplified the `if` a bit.


janpio:
Original approach of making it fully dynamic does not work because of https://docs.github.com/en/actions/learn-github-actions/contexts#github-context:~:text=the%20current%20job.-,Note%3A%20This%20context%20property%20is%20set%20by%20the%20Actions%20runner%2C%20and%20is%20only%20available%20within%20the%20execution%20steps%20of%20a%20job.%20Otherwise%2C%20the%20value%20of%20this%20property%20will%20be%20null.,-github.path

> Note: This context property is set by the Actions runner, and is only available within the execution steps of a job. Otherwise, the value of this property will be null.

.github/workflows/test.yaml
```diff
@@ -177,7 +192,7 @@ jobs:
 
   process-managers:
     needs: [detect_jobs_to_run]
-    if: ${{ fromJson(needs.detect_jobs_to_run.outputs.jobs)['process-managers'] == true }}
+    if: contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), github.job)
```
```suggestion
    if: contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), 'process-managers')
```

.github/workflows/test.yaml
```diff
@@ -74,7 +74,22 @@ jobs:
       - name: Install Dependencies
         run: pnpm install
       - id: detect
-        run: ./.github/workflows/detect-jobs-to-run.js <<<'${{ steps.files.outputs.all }}'
+        run: ./.github/workflows/scripts/detect-jobs-to-run.js <<<'${{ steps.files.outputs.all }}'
+
+  debug_detect_jobs_to_run:
+    needs: [detect_jobs_to_run]
+    name: Debug detected jobs to run  
+    if: ${{ contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), github.job) }}
```
```suggestion
    if: ${{ contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), 'detect_jobs_to_run') }}
```

.github/workflows/scripts/detect-jobs-to-run.test.js
```diff
@@ -0,0 +1,254 @@
...
+
+
+
+
+
+
+
+
+
+
```
janpio:
```suggestion
```

.github/workflows/scripts/detect-jobs-to-run.test.js
```diff
@@ -0,0 +1,254 @@
...
```
janpio:
```suggestion
```

.github/workflows/scripts/detect-jobs-to-run.test.js
```diff
@@ -0,0 +1,254 @@
...
```
janpio:
```suggestion
```

Jolg42 (approved):
Looks great! Thanks!

Jolg42:
Note: I was curious to check and everything looks good in this PR https://github.com/prisma/ecosystem-tests/pull/4315

PR #4345, opened 2023-12-06
test: add AWS Lambda `nodejs20.x`

Jolg42:
Announcement: https://aws.amazon.com/blogs/compute/node-js-20-x-runtime-now-available-in-aws-lambda/


platforms-serverless/lambda-node-16/run.sh
```diff
@@ -0,0 +1,37 @@
...
+pnpm install
+pnpm prisma generate
+pnpm tsc
+
+rm -rf lambda.zip
+zip --symlinks -r lambda.zip index.js prisma/schema.prisma node_modules/@prisma/client node_modules/.pnpm/@prisma+client*
+du -b ./lambda.zip
+
+AWS_RUNTIME=nodejs20.x
+AWS_RUNTIME_VERSION=20
```
janpio: This does not match the test/project name - intentional?

Jolg42: No, I already fixed that locally

Jolg42:
The functions were created 🎊 
https://eu-central-1.console.aws.amazon.com/lambda/home?region=eu-central-1#/functions?fo=and&o0=%3A&v0=platforms-serverless-lambda
Screenshot 2023-12-06 at 13 56 36

platforms-serverless/lambda-node-16/run.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+set -eux
```
jkomyno, line 3:
```suggestion
set -euox pipefail
```

platforms-serverless/lambda-node-20/run.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+set -eux
```
jkomyno, line 3: With a small adjustment, this should ensure that we return a non-zero exit code if the script breaks halfway:

```suggestion
set -euox pipefail
```

jkomyno (approved):
Approved with two minor comments

platforms-serverless/lambda-node-16/run.sh
Jolg42: Could you make this an issue maybe?

Because I don't think this should be done in this PR, if it's useful we could actually do it in all the bash scripts we have.

PR #4371, opened 2023-12-09
fix: Do not fallback to all on small workflow changes

janpio:


janpio:
This could help: https://stackoverflow.com/a/66921066

The GH actions only return the list of changed files: 
- https://github.com/marketplace/actions/changed-files#output_modified_files
- https://dev.to/scienta/get-changed-files-in-github-actions-1p36

Issue #1517, opened 2021-03-16
Conditional build + installation of Puppeteer is only partially working


janpio:
In theory:
https://github.com/prisma/e2e-tests/blob/39328785e1f33b0f1f37540336049b7618cab59d/.github/workflows/test.yaml#L407-L412

In practice:
![image](https://user-images.githubusercontent.com/183673/111310982-fa0a6d80-865d-11eb-8a6e-f36671f5e8e2.png)

janpio:
Turns out this is only the build of the container - the actual installation is indeed skipped as it should. But the installation is super fast anyway, it's the build that takes time and resources :/ So we better split out the codesandbox test to optimize this.

Jolg42:
Note: random find there is also this issue about improving codesandbox test https://github.com/prisma/ecosystem-tests/issues/1795

Jolg42:
Still hapening as of today

PR #4389, opened 2023-12-15
test(optional): add m1-office & move codesandbox to optional

Jolg42:
Closes: [#756](https://github.com/prisma/team-orm/issues/756)

It adds a new test, `m1-office` for testing on a macOS M1 machine located in Prisma's office.
We already have a test for this `m1-macstadium`.

If `m1-office` works well, `m1-macstadium` will be removed later.

Also by moving Codesandbox in its own job it closes
https://github.com/prisma/ecosystem-tests/issues/1517
I also moved it to `optional` tests.


apolanc (approved):

.github/scripts/test-project.sh
```diff
@@ -6,6 +6,12 @@ shopt -s inherit_errexit || true
 export CI=true
 
 dir=$1
+
+if [[ "$dir" = "platforms-codesandbox" || "$dir" = "platforms-m1-office" ]]
```
Jolg42, line 14:
```suggestion
# These are in optional-test.yaml, under the `platforms` directory
# But the names don't match the directory name
# Example `platforms-m1-office` is in `platforms` directory
# So this renames the key so it's correctly processed
if [[ "$dir" = "platforms-codesandbox" || "$dir" = "platforms-m1-office" ]]
```

.github/scripts/check-folders.js
```diff
@@ -83,17 +84,27 @@ async function main() {
...
+          return acc.concat(...matrix_optional[key])
+        }, [])
+
+      // These are in optional-test.yaml, under the `platforms` directory
+      // But the names don't match the directory name
+      // Example `platforms-m1-office` is in `platforms` directory
+      // So this renames the key so it's correctly processed
+      if (key === 'platforms-m1-office' || key === 'platforms-codesandbox') {
+        key = 'platforms'
+      }
```
Jolg42, line 103: Current hack to make the checks-folders pass without introducing a new top-level directory

I suggest we keep the hack for now and merge, and I'm open to suggestion to clean this up in a follow-up PR.

.github/scripts/test-project.sh
```diff
@@ -6,6 +6,16 @@ shopt -s inherit_errexit || true
 export CI=true
 
 dir=$1
+
+# These are in optional-test.yaml, under the `platforms` directory
+# But the names don't match the directory name
+# Example `platforms-m1-office` is in `platforms` directory
+# So this renames the key so it's correctly processed
+if [[ "$dir" = "platforms-codesandbox" || "$dir" = "platforms-m1-office" ]]
```
Jolg42, line 14: Current hack to make the script work without introducing a new top-level directory

.github/scripts/check-engines-cli.sh
```diff
@@ -21,6 +21,7 @@ fi
 skipped_projects=(
   aws-graviton                    # No local project at all (everything happens on server), so no `prisma` or `node_modules
   firebase-functions              # No local project at expected location (but in `functions` subfolder)
+  m1-office                       # No local project at all (everything happens on server), so no `prisma` or `node_modules
```
millsp:
```suggestion
  m1-office                       # No local project at all (everything happens on server), so no `prisma` or `node_modules`
```

platforms/aws-graviton/run.sh
```diff
@@ -4,30 +4,20 @@ set -eux
...
-      rm -rf ./node_modules;
-      npm i -g pnpm@8;
-      pnpm install;
-      pnpm prisma generate;
-      pnpm prisma -v;
-  "
-fi
+echo "PRISMA_CLIENT_ENGINE_TYPE = $PRISMA_CLIENT_ENGINE_TYPE"
 
+ssh -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com -tt "
```
millsp, line 9: Nice!

platforms/m1-office/code/m1.js
```diff
@@ -0,0 +1,19 @@
...
+const isArm = uname === "arm64";
+
+let isM1 = false;
+try {
+    // this command will only succeed on m1 macs.
+    execSync(`arch -arm64e echo hi`);
+    isM1 = true;
+} catch (err) {
+    // Must not be an m1 mac
+}
```
millsp: Should we let this be a plain error?

platforms/m1-office/finally.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eux
+
+MACHINE_IP=192.168.1.192
```
millsp, line 5: I guess the machine has a fixed ip configured?

platforms/m1-office/run.sh
```diff
@@ -0,0 +1,44 @@
...
+        nvm -v;
+        node -v;
+        npm -v;
+    else
+        echo 'Node.js will be isntalled...';
+        curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash;
+        nvm install 20;
+        nvm alias default 20;
+        nvm ls;
+    fi
```
millsp: Is this still necessary?

platforms/m1-office/run.sh
millsp, line 1: Nice one!

platforms/m1-office/finally.sh
Jolg42: Indeed, the internal IP is fixed

.github/workflows/optional-test.yaml
```diff
@@ -147,3 +147,140 @@ jobs:
...
+  platforms-m1-office:
+    needs: [detect_jobs_to_run]
+    if: contains(fromJSON(needs.detect_jobs_to_run.outputs.jobs), 'platforms')
+
+    timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - m1-office
```
millsp, line 219: Is this required?

.github/scripts/check-folders.js
millsp, line 103: Could we introduce a top-level folder for `platforms` instead?

.github/scripts/check-engines-cli.sh
```diff
@@ -21,6 +21,7 @@ fi
...
   m1-macstadium                   # No local project at all (everything happens on server), so no `prisma` or `node_modules
```
Jolg42:
```suggestion
  aws-graviton                    # No local project at all (everything happens on server), so no `prisma` or `node_modules`
  firebase-functions              # No local project at expected location (but in `functions` subfolder)
  m1-office                       # No local project at all (everything happens on server), so no `prisma` or `node_modules`
  m1-macstadium                   # No local project at all (everything happens on server), so no `prisma` or `node_modules`
```

.github/scripts/check-engines-client.sh
```diff
@@ -20,6 +20,7 @@ skipped_projects=(
   # no local project
   aws-graviton                                # No local project at all (everything happens on server), so no `prisma` or `node_modules
   m1-macstadium                               # No local project at all (everything happens on server), so no `prisma` or `node_modules
+  m1-office                                   # No local project at all (everything happens on server), so no `prisma` or `node_modules
```
Jolg42:
```suggestion
  aws-graviton                                # No local project at all (everything happens on server), so no `prisma` or `node_modules`
  m1-macstadium                               # No local project at all (everything happens on server), so no `prisma` or `node_modules`
  m1-office                                   # No local project at all (everything happens on server), so no `prisma` or `node_modules`
```

.github/scripts/check-folders.js
Jolg42: We already have a top folder named `platforms`

.github/workflows/optional-test.yaml
Jolg42: No it's from copy paste, but it makes it easier to add another test if ever needed in the job

.github/scripts/check-folders.js
millsp: Ok, then I didn't understand the hack 😅 Would you mind explaining it to me again?

platforms/m1-office/code/m1.js
Jolg42: It's a copy paste but a good time to improve this I think!

Jolg42: Done!

platforms/m1-office/run.sh
```diff
@@ -0,0 +1,44 @@
...
+    which pnpm;
+
+    if which node > /dev/null
+    then
+        echo 'Node.js is already installed';
+        nvm -v;
+        node -v;
+        npm -v;
+    else
+        echo 'Node.js will be isntalled...';
```
Jolg42:
```suggestion
        echo 'Node.js will be installed...';
```

platforms/m1-office/run.sh
Jolg42: What do you mean?
I think we can keep this, it would only run if Node.js is not detected.

.github/scripts/check-folders.js
Jolg42: I also don't like it that much but would prefer to change that later.
Currently, the setup expects to have a 1:1 with top level directories and job names in the Actions yaml.

So the names `platforms-m1-office` and `platforms-codesandbox` are actually defined in `optional-test.yaml` but since the name is different from the directory tree it would fail the expectations.

I made codesandbox a job of its own because the test requires Puppeteer and there is no way to make the setup of Puppeteer conditional, which makes all the tests not using Puppeteer slower.

platforms/m1-office/run.sh
millsp: I assumed the machine was already setup

.github/scripts/check-folders.js
millsp: Ah ok, gotcha, thanks for the explanation. If I understood correctly, then I'd prefer having the projects being like
- `platforms/platforms-m1-office`
- `platforms/platforms-codesandbox`

No strong opinion though :P

millsp (approved):

platforms/m1-office/run.sh
Jolg42: It also makes it easier to know and change the setup later

.github/scripts/check-folders.js
Jolg42: That was not an option since I wanted to have codesandbox and the m1 test running separately in a different job 
not sure if that's really worth it in optional yaml at this point
I'll move them together again later in optional yaml to simplify things!

PR #4440, opened 2024-01-08
chore(azure functions): stop functions after test failed

Jolg42:
Stopping the function makes the infrastructure cost predictable, as it cannot be abused intentionally or not. (Functions are normally free since the volume of requests is very low).

It might also be faster to delete them later?

### DIFF
--- a/platforms-serverless/azure-functions-linux/finally.sh
+++ b/platforms-serverless/azure-functions-linux/finally.sh
@@ -28,5 +28,6 @@ else
 
     # stop the function
     # Note that re-running the test will create a new function from scratch
+    # so this does not have negative effect on any re-run
     az functionapp stop --name "$app" --resource-group "$group"
 fi
--- a/platforms-serverless/azure-functions-linux/finally.sh
+++ b/platforms-serverless/azure-functions-linux/finally.sh
@@ -18,7 +18,6 @@ if [ "$TEST_EXIT_CODE" -eq "0" ]; then
 
     az config set extension.use_dynamic_install=yes_without_prompt
     az monitor app-insights component delete --app "$app" --resource-group "$group"
-
 else
     # If tests failed
     echo "To view Logs visit"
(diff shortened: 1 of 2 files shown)


platforms-serverless/azure-functions-linux/finally.sh
```diff
@@ -27,4 +26,7 @@ else
     # move function code back to placeholder location, so we can try again
     mv "$app" "func-placeholder"
 
+    # stop the function
+    # Note that re-running the test will create a new function from scratch
+    az functionapp stop --name "$app" --resource-group "$group"
```
janpio, line 32: What does "stopping" a function really do? What happens if the function is not stopped?

Jolg42: When the function is running, it's listening for HTTP requests, when it's stopped, it's not listening.

platforms-serverless/azure-functions-linux/finally.sh
```diff
@@ -27,4 +26,7 @@ else
     # move function code back to placeholder location, so we can try again
     mv "$app" "func-placeholder"
 
+    # stop the function
+    # Note that re-running the test will create a new function from scratch
```
janpio, line 30:
```suggestion
    # Note that re-running the test will create a new function from scratch
    # so this does not have negative effect on any re-run
```

platforms-serverless/azure-functions-linux/finally.sh
janpio: Ah, get it. Suggestion: https://github.com/prisma/ecosystem-tests/pull/4440/files#r1446076654

platforms-serverless/azure-functions-windows/finally.sh
```diff
@@ -27,4 +26,7 @@ else
     # move function code back to placeholder location, so we can try again
     mv "$app" "func-placeholder"
 
-fi
\ No newline at end of file
+    # stop the function
+    # Note that re-running the test will create a new function from scratch
```
janpio, line 30:
```suggestion
    # Note that re-running the test will create a new function from scratch
    # so this does not have negative effect on any re-run
```

PR #4454, opened 2024-01-11
chore: rename pdp tests to accelerate

### LINKED ISSUE #4378: Remove Data Proxy (1) tests and rename tests to Accelerate
When Data Proxy is sunset in the new year, we can remove all the existing tests.
We should also renamed the whole folder and setup to `Accelerate` to be less confusing.

millsp:
Renames the pdp tests to Accelerate, does not rename all `DATAPROXY` namings in env vars.

Closes https://github.com/prisma/ecosystem-tests/issues/4378


Jolg42 (approved):
✨

PR #4489, opened 2024-01-23
test(driver-adapters): custom output

millsp:
Uses the shell wand :magic_wand:  to test all our driver adapter tests with a custom output and configured as documented. This means we now have two steps instead of one, to test with regular and custom outputs within the same project.

Failing tests are [unrelated](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1706044899165809).

closes https://github.com/prisma/team-orm/issues/862

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -67,6 +67,17 @@ fi
 echo "$ cd $dir/$project"
 cd "$dir/$project"
 
+# if FORCE_CUSTOM_OUTPUT is set, we execute 3 commands that will turn the project into a custom output project
+if [ -n "$FORCE_CUSTOM_OUTPUT" ]; then
+  echo "-----------------------------"
+  echo ""
+  echo "FORCE_CUSTOM_OUTPUT=$FORCE_CUSTOM_OUTPUT, executing 3 commands to turn the project into a custom output project"
+  echo ""
+  find . -name "*.js" | grep -v node_modules | grep -v test | xargs sed -i "s/@prisma\/client/db/g"
+  find . -name "*.prisma" | grep -v node_modules | xargs sed -i '/generator client {/a output = "client"'
+  find . -name "package.json" | grep -v node_modules | xargs sed -i '/"dependencies": {/a "db": "link:./prisma/client",'
+fi
+
 if [ -f "prepare.sh" ]; then
   echo "-----------------------------"
   echo ""
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -74,9 +74,18 @@ if [ -n "${FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT+x}" ]; then
   echo "FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT=$FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT, executing commands to turn the project into a custom output project"
   echo ""
   find . -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 sed -i "s/@prisma\/client/db/g"
-  find . -name "*.prisma" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i '/generator client {/a output = "client"'
-  find . -name "package.json" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i '/"dependencies": {/a "db": "link:./prisma/client",'
+  find . -name "*.prisma" | grep -v node_modules | xargs sed -i 's/provider\s*=\s*"prisma-client-js"/&\noutput = "client"/g'
+  find . -name "package.json" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i 's/"dependencies": {/&\n"db": "link:.\/prisma\/client",/g'
   bash ../../scripts/update-locks.sh
+  
+  # this adds some level of safety to ensure that js files were at least modified as one would expect
+  if [[ $(git status --porcelain | grep "$dir/$project" | grep -c ".js$") -gt 0 ]];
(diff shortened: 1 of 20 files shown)


.github/scripts/test-project.sh
```diff
@@ -67,6 +67,18 @@ fi
...
+  echo "-----------------------------"
+  echo ""
+  echo "FORCE_CUSTOM_OUTPUT=$FORCE_CUSTOM_OUTPUT, executing commands to turn the project into a custom output project"
+  echo ""
+  find . -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 sed -i "s/@prisma\/client/db/g"
+  find . -name "*.prisma" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i '/generator client {/a output = "client"'
+  find . -name "package.json" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i '/"dependencies": {/a "db": "link:./prisma/client",'
+  bash ../../scripts/update-locks.sh
+fi
+
```
millsp: The bulk of the change.

driver-adapters-wasm/neon-cfpages-basic/prisma/schema.prisma
```diff
@@ -1,7 +1,7 @@
 generator client {
   provider        = "prisma-client-js"
   previewFeatures = ["driverAdapters"]
-  engineType      = "wasm"
```
millsp, line 4: Forgotten cleanup from a previous PR.

.github/scripts/test-project.sh
```diff
@@ -67,6 +67,18 @@ fi
 echo "$ cd $dir/$project"
 cd "$dir/$project"
 
+# if FORCE_CUSTOM_OUTPUT is set, we execute commands that will turn the project into a custom output project
```
Jolg42:
```suggestion
# if FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT is set, we execute commands that will turn the project into a custom output project
```
Could you rename it to something like this, maybe?

`FORCE_CUSTOM_OUTPUT` is a bit too generic I think, it could be about many things.

Jolg42 (commented):
I think the `concurrency` in the `test.yaml` needs to be adapted for `driver-adapters` and `driver-adapters-wasm` because this will duplicate tests and the hosted infrastructure will randomly fail because it's not isolated per test.

.github/scripts/test-project.sh
```diff
@@ -67,6 +67,18 @@ fi
...
+if [ -n "${FORCE_CUSTOM_OUTPUT+x}" ]; then
+  echo "-----------------------------"
+  echo ""
+  echo "FORCE_CUSTOM_OUTPUT=$FORCE_CUSTOM_OUTPUT, executing commands to turn the project into a custom output project"
+  echo ""
+  find . -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 sed -i "s/@prisma\/client/db/g"
+  find . -name "*.prisma" ! -path "*/node_modules/*" -print0 | xargs -0 sed -i '/generator client {/a output = "client"'
```
Jolg42: I'm not a big fan of coupling the generator name here but at the same time it looks ok at the moment, it's what we use in all tests.

millsp: You're right, I should not, I could match on the generator provider instead.

Jolg42:
@millsp What do you think about the concurrency? https://github.com/prisma/ecosystem-tests/pull/4489#pullrequestreview-1840801531

millsp:
They should have independent database urls already, so I guess you mean this is needed for some of the deployments?

janpio:
Do I understand correctly that this adds a second _step_ for both jobs that, after running them with no custom output now also runs it via a custom output?

Then `concurrency` should not be impacted as that is per _job_ - which now just takes longer and uses the same database twice.

.github/scripts/test-project.sh
```diff
@@ -67,6 +67,28 @@ fi
 echo "$ cd $dir/$project"
 cd "$dir/$project"
 
+# if FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT is set, we execute commands that will turn the project into a custom output project
```
janpio, line 70: This would have been worth a comment on _how_ we do that, took me a bit to understand the commands below.

millsp: Here's a summary of what it does

```sh
○ find is used to list files.
  ◆ -name "package.json" stipulates that we search for files named package.json.
  ◆ ! -path "*/node_modules/*" stipulates that we exclude all files in the node_modules folder.
  ◆ -print0 specifies that we want to print the list of files separated by the null character.
○ xargs is used to execute a command for each item in a list.
  ◆ -0 specifies that the list is separated by the null character.
  ◆ -r specifies that we only execute the command if the list is not empty.
○ sed is used to search and replace text in files.
  ◆ -i specifies that we want to replace the text in the file.
  ◆ 's/"dependencies": {/&\n"db": "link:.\/prisma\/client",/g' is the search and replace pattern.
    ◇ s means we want to search and replace.
    ◇ "dependencies": { is the text we want to search for.
    ◇ & means we want to replace it with itself.
    ◇ \n means we want to insert a newline.
    ◇ "db": "link:.\/prisma\/client", is the text we want to insert.
    ◇ /g means we want to replace all occurrences.
 ```

**Comment by Jolg42** [comment]
The current concurrency was I think already problematic 
```
concurrency: ${{ github.job }}-${{ matrix.combo }}-${{ matrix.clientEngine }}
```
And adding one more step doesn't make it better.

I guess all of this is hidden in the retries' mechanism, since if anything fails, a retry will happen, let's see how this goes.

PR #4493, opened 2024-01-25
fix(driver-adapters-wasm): cleanup

millsp:
My main intention was to close https://github.com/prisma/team-orm/issues/764, which was done mostly via the Cloudflare UI and also I did increase the wait period of a page deployment to reduce flakiness already before. So this is just a cleanup/fix/improvement for things I found along the way.


driver-adapters-wasm/neon-node-nextjs-edgemw/run.sh
```diff
@@ -12,5 +12,6 @@ wait_for_host() {
 
 pnpm install
 pnpm next dev --port 3000 &
+echo $! > server.pid
```
millsp, line 16: Caused the second step for custom output tests to fail.

driver-adapters-wasm/pg-cfpages-basic/functions/function.js
```diff
@@ -2,9 +2,10 @@
 import { Prisma, PrismaClient } from '@prisma/client'
 import { Pool } from 'pg'
 import { PrismaPg } from '@prisma/adapter-pg'
+import { DATABASE_URL } from './dbUrl.js'
 
 export async function onRequest(context) {
-  const client = new Pool({ connectionString: context.env.DATABASE_URL })
+  const client = new Pool({ connectionString: DATABASE_URL })
```
millsp, line 9: Now uses the postgresql database provided by ecosytem-tests instead.

driver-adapters-wasm/neon-node-nextjs-edgemw/run.sh
Jolg42: What do you mean? It's hard to understand without context for me at the moment 🤔

Jolg42, line 16:
```suggestion
# create a `server.pid` file, which will be used in `finally.sh` to kill the process
echo $! > server.pid
```

Jolg42 (approved):

driver-adapters-wasm/pg-cfpages-basic/functions/function.js
Jolg42, line 9:
```suggestion
	// `DATABASE_URL` comes from `dbUrl.js` and its value is set by `prepare.sh`
  const client = new Pool({ connectionString: DATABASE_URL })
```

PR #4501, opened 2024-01-30
tests: add D1 on CF + CF Pages

Jolg42:
Closes https://github.com/prisma/team-orm/issues/899


driver-adapters-wasm/d1-cf-basic/src/index.js
```diff
@@ -0,0 +1,190 @@
...
+        //     name: 'Test 4',
+        //   },
+        //   update: {},
+        //   select: {
+        //     email: true,
+        //     age: true,
+        //     name: true,
+        //   },
+        // }),
+      }
```
millsp: I guess this can be enabled since it's not an http resquest?

millsp (approved):
Looks good! (+see comment)

driver-adapters-wasm/d1-cf-basic/src/index.js
Jolg42: I put that back, I wanted to but forgot at some point, thanks for mentioning it!

PR #4574, opened 2024-02-19
fix Cloudflare Pages integration

petebacondarwin:
There is a bug in `wrangler pages deploy` where it will still try to compile the `functions` directory even if the `_worker.js` directory exists. The steps in this PR should work around that (and the fact that we cannot specify `--node-compat` during `wrangler pages deploy`.


janpio:
Thanks!

We'll try to make this branch local so that the secrets are provided and we can try this out.

janpio:
@petebacondarwin That partially worked: https://github.com/prisma/ecosystem-tests/pull/4575 (The remaining failure could be because we run the test twice in slightly different configurations)

millsp:
Hey @petebacondarwin thank you so much for the PR and the workaround. We moved your commits into a mirror [PR](https://github.com/prisma/ecosystem-tests/pull/4575) to run the tests, fix our script, and apply some minor cosmetic changes. Thanks for taking the time to contribute.

petebacondarwin:
Glad I could help.

PR #4612, opened 2024-03-01
chore(deps): jest major

millsp:
The core of the change for the update is to of course bump `jest` and `ts-jest`, but also change snapshots from `Object { ...` to `{ ...` and `Array [...` to `[...`.


Jolg42:
Nice one 🙌🏼

PR #4613, opened 2024-03-01
test(driver-adapters): imports map

millsp:
For the review, use this [filter](https://github.com/prisma/ecosystem-tests/pull/4613/files?file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&show-viewed-files=true)

### DIFF
--- a/.github/scripts/test-project.sh
+++ b/.github/scripts/test-project.sh
@@ -74,11 +74,9 @@ if [ -n "${FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT+x}" ]; then
   echo ""
   echo "FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT=$FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT, executing commands to turn the project into a custom output project"
   echo ""
-  find . -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/db/g"
-  find . -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/db/g"
+  find . -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/prisma\/client/g"
+  find . -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/prisma\/client/g"
   find . -type f -name '*.prisma' ! -path '*/node_modules/*' -print0 | xargs -0 -r sed -i 's/provider\s*=\s*"prisma-client-js"/&\noutput = "client"/g'
-  find . -type f -name "package.json" ! -path "*/node_modules/*" -print0 | xargs -0 -r sed -i 's/"dependencies": {/&\n"db": "link:.\/prisma\/client",/g'
-  bash ../../scripts/update-locks.sh
 
   # this adds some level of safety to ensure that js files were at least modified as one would expect
   if [[ $(git status --porcelain | grep "$dir/$project" | grep -Ec ".m?js$") -gt 0 ]];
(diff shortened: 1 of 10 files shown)


.github/scripts/test-project.sh
```diff
@@ -73,12 +73,18 @@ if [ -n "${FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT+x}" ]; then
...
+  find . -mindepth 4 -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/\.\.\/\.\.\/prisma\/client/g"
+  find . -mindepth 3 -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/\.\.\/prisma\/client/g"
+  find . -mindepth 3 -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/\.\.\/prisma\/client/g"
+  find . -mindepth 2 -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/prisma\/client/g"
+  find . -mindepth 2 -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\.\/prisma\/client/g"
+  find . -mindepth 1 -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\/prisma\/client/g"
+  find . -mindepth 1 -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/\.\/prisma\/client/g"
```
millsp: Ugly buuut understandable and easily scaled if need be. Alternatively, we could make it a loop but it would  be harder to understand imo, something along these lines:

```sh
  for depth in {4..1}; do
    find . -mindepth $depth -type f -name "*.js" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/$(printf '../%.0s' $(seq 1 $((depth-1))))prisma\/client/g"
    find . -mindepth $depth -type f -name "*.mjs" ! -path "*/node_modules/*" ! -name "*test*" -print0 | xargs -0 -r sed -i "s/@prisma\/client/$(printf '../%.0s' $(seq 1 $((depth-1))))prisma\/client/g"
  done
  ```

File: .github/scripts/test-project.sh [hunk 2]
```diff
@@ -73,12 +73,18 @@ if [ -n "${FORCE_PRISMA_CLIENT_CUSTOM_OUTPUT+x}" ]; then
...
+
```
millsp: So above: it is the DX improvement, custom paths are back and work on workers!

driver-adapters-wasm/neon-http-vercel-nextjs-edgemw/next.config.js
```diff
@@ -1,14 +1,6 @@
...
-const nextConfig = {
-  webpack(config) {
-    config.experiments ??= {}
-
-    config.experiments.asyncWebAssembly = true
-
-    return config
-  }
-}
+const nextConfig = {}
```
millsp, line 4: And then: this is the other DX improvement, where we can remove ugly configs!

.github/scripts/test-project.sh
Druue: I think I find the loop easier to read honestly. They're still both rough on the line length tho, is it possible to separate them out over multiple? e.g.

```sh
find . -mindepth $depth -type f -name "*.js" 
    ! -path "*/node_modules/*" 
    ! -name "*test*" -print0 
    | xargs -0 -r sed -i "s/@prisma\/client/$(printf '../%.0s' $(seq 1 $((depth-1))))prisma\/client/g"
```

janpio:
Restarted test `platforms-serverless-vercel (vercel-with-nextjs-caching, binary, ubuntu-20.04)` (because of database reset failure, probably some concurrency problem)

Update: That fixed it.

jkomyno (approved):
🚀

.github/prisma-version.txt
```diff
@@ -1 +1 @@
-5.11.0-dev.19
+5.11.0-integration-fix-client-wasm-loading-vercel.10
```
Jolg42: You need to change this, I think?

.github/scripts/test-project.sh
SevInf: I think at this point we might just as well use a proper programming language instead of chain shell commands

SevInf: did just that

PR #4769, opened 2024-04-04
ci(azure-functions): add draft to delete old resources

Jolg42:
[skip ci]

PR #4770, opened 2024-04-04
ci(az delete): add delete step with retries

Jolg42:
Follows https://github.com/prisma/ecosystem-tests/pull/4769


Jolg42:
See run for `-2024-01-`
https://github.com/prisma/ecosystem-tests/actions/runs/8555209106/job/23442121370

PR #4914, opened 2024-05-07
test(turso): uncomment orThrow methods

Jolg42:
Closes https://github.com/prisma/team-orm/issues/663

It turns out that it just works.


Jolg42:
Note from @janpio we could have a test where `orThrow` throws, I created https://github.com/prisma/ecosystem-tests/issues/4918

millsp (approved):

PR #4915, opened 2024-05-07
test(pg-cf-hyperdrive): improve delete hyperdrive script

Jolg42:
It's failing on the default branch `dev`, it looks like the script cannot recover if the hyperdrive config does not exist because it will try to delete it and fail.

This adds a check that if it does not exist, it continues.


Jolg42:
It works ✅ 
https://github.com/prisma/ecosystem-tests/actions/runs/8985295019/job/24679012209?pr=4915#step:9:125

Jolg42:
Merging to bring back the suite into a passing state.

PR #4917, opened 2024-05-07
test(sqlite): add `.createManyAndReturn()`

Jolg42:
adds `.createManyAndReturn()` in current tests

Related to https://github.com/prisma/team-orm/issues/1107


driver-adapters-wasm/d1-cf-basic/src/index.js
```diff
@@ -111,6 +135,11 @@ export default {
           _count: {
             age: true,
           },
+          orderBy: {
+            _count: {
+              age: 'asc',
+            },
+          },
```
janpio, line 142: Needed for stable ordering?

Jolg42: Yes!

driver-adapters/planetscale-lambda-basic/index.js
```diff
@@ -40,6 +40,30 @@ exports.handler = async () => {
           },
         ],
       }),
+      // Not implemented for MySQL
```
SevInf: I think we can remove this completely for now: as far as we concerned, with current state of things this is not a temporary limitation, `createManyAndReturn` can not be implemented for mysql at all and very likely we'll never need this test ate all.

driver-adapters/planetscale-lambda-basic/index.test.js
```diff
@@ -37,6 +37,26 @@ test('prisma version and output', async () => {
 }
 `)
   expect(regResult.createMany.count).toBe(2)
+  // Not implemented for MySQL
+  // expect(regResult.createManyAndReturn).toMatchInlineSnapshot(`
```
SevInf, line 41: Same

driver-adapters/neon-http-lambda-basic/index.js
```diff
@@ -40,6 +40,29 @@ exports.handler = async () => {
     //     },
     //   ],
     // }),
+    // createManyAndReturn: await prisma.user.createManyAndReturn({
```
Druue, line 44: I imagine this is commented out because http mode but it would be nice to have a comment here about it

SevInf (approved):
I think we should remove commented out MySQL tests - it is not a temporary limitation. `createManyAndReturn` can not be implemented for MySQL, so there is little value for keeping them around. 

But I don't want to block it further, so approving now and trust you to do the changes.

Druue (approved):

driver-adapters/neon-http-lambda-basic/index.js
Jolg42: It's the same comment as above normally, so I did not add anything
But it looks like the previous block does not have a comment for this block here, it's a bit of a mess these tests

PR #5037, opened 2024-06-03
fix(studio): Ask hash from Studio itself instead of computing it

SevInf:
Way we compute hash have changed slightly after multi-file work: schemas
always pass through `mergeSchemas` call which also does reformatting.

Intstead of trying to duplicate this logic in our tests to ensure the
hash matches we now just ask for schema hash from a studio itself -
this would make the test more resilient to such changes.


Jolg42 (approved):
Nice!

Druue (approved):

PR #5039, opened 2024-06-04
feat(driver-adapters): pglite-prisma-adapter

janpio:
Test using https://github.com/lucasthevenet/pglite-utils/tree/main/packages/prisma-adapter


driver-adapters/pglite-node-basic/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "pglite-node-basic",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "test": "jest index.test.js"
```
Jolg42:
```suggestion
    "test": "jest"
```
Should be enough

janpio: Gives me this:
```
gitpod /workspace/ecosystem-tests/driver-adapters/pglite-node-basic (dev) $ pnpm jest index.test.ts
 FAIL  ./index.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /workspace/ecosystem-tests/driver-adapters/pglite-node-basic/node_modules/.pnpm/@electric-sql+pglite@0.1.5/node_modules/@electric-sql/pglite/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import{a,b,c,d,e,f}from"./chunk-OVRU7FYL.js";import"./chunk-EKNQE2HU.js";export{a as Mutex,d as PGlite,e as messages,c as parse,f as protocol,b as types};
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import { Prisma, PrismaClient } from '@prisma/client'
    > 2 | import { PGlite } from '@electric-sql/pglite'
        | ^
      3 | import { readFile } from 'fs/promises'
      4 | import { PrismaPGlite } from 'pglite-prisma-adapter'
      5 |

      at Runtime.createScriptFromCode (node_modules/.pnpm/jest-runtime@29.7.0/node_modules/jest-runtime/build/index.js:1505:14)
      at Object.<anonymous> (index.ts:2:1)
      at Object.<anonymous> (index.test.ts:1:1)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        1.896 s
Ran all test suites matching /index.test.ts/i.
```

(But same with just `pnpm jest` honestly running manually)

Jolg42 (approved):

PR #5074, opened 2024-06-10
pkg: Remove useless generate

SevInf:
This `generate` call is not supposed to work - there is no schema to
generate from where it is called. It was working before because generate
incorrectly thought it's called from a postinstall hook and produced
soft error instead.

Close prisma/team-orm#1183

### DIFF
--- a/binaries/pkg/run.sh
+++ b/binaries/pkg/run.sh
@@ -3,7 +3,6 @@
 set -eu
 
 pnpm install
-pnpm prisma generate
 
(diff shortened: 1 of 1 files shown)


Druue (approved):

Druue:
This is pending the following PR, yeah?
https://github.com/prisma/prisma/pull/24484

Jolg42 (approved):

SevInf:
@Druue they are part of same bunch, but can be merged independently

PR #5177, opened 2024-07-05
fix: Firebase authentication

SevInf:
Probably got broken with Joël account deactivation. Switches to service account and modern authentication method.


platforms-serverless/firebase-functions/test.sh
```diff
@@ -16,7 +16,7 @@ expected='{"version":"'$prisma_version'","createUser":{"id":"12345","email":"ali
 actual=$(curl "$url")
 
 # TODO Why is this here?
```
Druue: 🤔

SevInf: Removed them, logs exist in finally too

jkomyno (approved):

PR #5196, opened 2024-07-08
fix: heroky-buildpack test

SevInf:
Looks like Heroku updated pnpm to not run post-install hooks by default.
Adding explicit build step.


jkomyno (approved):
Q: Does this mean that the `postinstall` hook stopped working on Heroku?

A: Yes, that seems to be the default behavior in newer `pnpm` versions

PR #5197, opened 2024-07-08
fix: Run yarn tests on node 18

SevInf:
They no longer work with node 16


jkomyno (approved):

jkomyno:
Any clue why these tests suddently got broken? Do we have a lax `yarn` dependency version anywhere?

SevInf:
It looks like we merge renovate PR regardless of engine failures, update happened and it was no longer compatible with 16

PR #5264, opened 2024-07-25
fix(aws-graviton): Cleanup pnpm packages after the test

SevInf:
We did not really clean up pnpm store after the test, so over the years it accumulated 27 GB worth of packages and the machine ran out of disk space.
Fixed by running `pnpm store prune` in `finally.sh` after removing test project from the machine.

Ideally, we'd just spin up new clean VM for every test, but I am not sure how that affects performance and costs, so doing simpler thing for now.

### DIFF
--- a/platforms/aws-graviton/finally.sh
+++ b/platforms/aws-graviton/finally.sh
@@ -2,5 +2,9 @@
 
 set -eux
 
-ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com "rm -rf /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID/$PRISMA_CLIENT_ENGINE_TYPE"
+ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com "
+    rm -rf /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID/$PRISMA_CLIENT_ENGINE_TYPE
+    pnpm cache prune
+"
+
 
--- a/platforms/aws-graviton/finally.sh
+++ b/platforms/aws-graviton/finally.sh
@@ -4,7 +4,9 @@ set -eux
 
 ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com "
     rm -rf /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID/$PRISMA_CLIENT_ENGINE_TYPE
-    pnpm cache prune
+    # delete parent folder if empty
+    find /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID -maxdepth 0 -empty -delete
+    pnpm store prune
 "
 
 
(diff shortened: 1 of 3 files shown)


platforms/aws-graviton/finally.sh
```diff
@@ -2,5 +2,11 @@
 
 set -eux
 
-ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com "rm -rf /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID/$PRISMA_CLIENT_ENGINE_TYPE"
+ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com "
+    rm -rf /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID/$PRISMA_CLIENT_ENGINE_TYPE
+    # delete parent folder if empty
+    find /home/ec2-user/aws-graviton/$GITHUB_JOB/$GITHUB_RUN_ID -maxdepth 0 -empty -delete
```
jkomyno, line 8: Why `-empty`?

platforms/aws-graviton/sync-from-remote.sh
```diff
@@ -9,11 +9,8 @@ echo "Sync: Syncing code from EC2"
 echo "Sync: Removing existing local code"
 rm -rf ./code
 
-echo "Sync: Removing remote node_modules on server"
-ssh -tt -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com 'rm -rf /home/ec2-user/aws-graviton/node_modules'
-
 echo "Sync: Copying remote code from EC2 to local"
-scp -i ./server-key.pem -rp ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com:/home/ec2-user/aws-graviton/ .
+rysnc -Pav -e "ssh -i ./server-key.pem" --exclude "node_modules" ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com:/home/ec2-user/aws-graviton/ .
```
jkomyno: I had to familiarise myself with this diff, here's what I got:

- Before, we securely transferred the current folder to EC2 via `scp`, specifying the private key file with `-i`, recursively copying folders (with `-r`) while preserving metadata such as access modes and time.
- Now, we use `rsync` - which is kind of an optimised `scp`, so perhaps it will lead to lower transferred traffic. `-P` keeps partially transferred files and shows progress interactively, `-a` is the same as `scp -rp`, and `-v` indicates "verbose mode". We're also skipping the transfer of the `node_modules` folder (with `--exclude`)

Sounds like a win to me, perhaps let's add a couple more comments to these scripts next time :)

platforms/aws-graviton/run.sh
```diff
@@ -12,8 +12,6 @@ ssh -i ./server-key.pem ec2-user@ec2-54-209-135-27.compute-1.amazonaws.com -tt "
     export CI=\"true\"
     export PRISMA_CLIENT_ENGINE_TYPE=\"$PRISMA_CLIENT_ENGINE_TYPE\"
 
-    rm -rf ./node_modules;
```
jkomyno, line 15: This is due to `rsync --exclude "node_modules"`, right?

jkomyno (approved):

platforms/aws-graviton/run.sh
SevInf: exactly, i wanted to preserve local `node_modules` so we could run `pnpm store prune` in finally. `pnpm` gets installed in `node_modules` and if we keep this line, it will be gone by the time we do a cleanup.

platforms/aws-graviton/sync-from-remote.sh
SevInf: `--exclude` option was actually main motivator for moving to `rsync`

SevInf: Replaced short `rsync` options with long ones, hope that helps.

PR #5424, opened 2024-09-03
feat(docker): add regression test for https://github.com/prisma/prisma/issues/25101

jkomyno:
This PR adds a Docker regression test for https://github.com/prisma/prisma/issues/25101.
Setup:
- Use Ubuntu 22.04
- Install Node.js 18.18.2
- Manually install `libssl.1.1`
- Run `prisma --version`
- The expected target is `debian-openssl-1.1.x`

See [Slack thread](https://prisma-company.slack.com/archives/C058VM009HT/p1725295332813769).

---

~~Btw, this reveals a bug in `@prisma/get-platform`, namely, in [`findLibSSLInLocations`](https://github.com/prisma/prisma/blob/0a423d475fd34bcea91918924352d9c12ac0ee37/packages/get-platform/src/getPlatform.ts#L405-L418). Prisma should return the "highest" LibSSL version in every specific path found (for Linux Ubuntu and other Debian-family distros, there are [two such paths](https://github.com/prisma/prisma/blob/0a423d475fd34bcea91918924352d9c12ac0ee37/packages/get-platform/src/getPlatform.ts#L309)), rather than stopping at the first `libssl` version encountered.~~

Update (after talking with @aqrln): the logic in `@prisma/get-platform`, although questionable, is intentional:
> it’s not a bug, it’s intentional behaviour (although one that I don’t necessarily agree with) that was introduced as a band-aid to work around some issue with some provider, i think netlify or something

### DIFF
--- a/docker/ubuntu-22.04-amd64-openssl-1.1.x/Dockerfile
+++ b/docker/ubuntu-22.04-amd64-openssl-1.1.x/Dockerfile
@@ -1,0 +1,62 @@
+# syntax=docker/dockerfile:1
+# Regression test for: https://github.com/prisma/prisma/issues/25101
+
+ARG UBUNTU_VERSION="22.04"
+
+FROM ubuntu:${UBUNTU_VERSION}
+
+ARG NODE_VERSION="18.18.2"
+ARG ARCH="x64"
+
+WORKDIR /usr/src/app
+COPY --from=app . ./
+COPY --from=utils . ./
+
+# Install system dependencies for Node.js
+RUN apt-get update -y \
+  && apt-get install -y curl xz-utils
+
+# Install Node.js
+RUN curl -fsSLO --compressed "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${ARCH}.tar.xz" \
+  && tar -xJf "node-v${NODE_VERSION}-linux-${ARCH}.tar.xz" -C /usr/local --strip-components=1 --no-same-owner \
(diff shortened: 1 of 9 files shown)


docker/ubuntu-22.04-amd64-openssl-1.1.x/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+
+# Install Node.js
+RUN curl -fsSLO --compressed "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${ARCH}.tar.xz" \
+  && tar -xJf "node-v${NODE_VERSION}-linux-${ARCH}.tar.xz" -C /usr/local --strip-components=1 --no-same-owner \
+  && rm "node-v${NODE_VERSION}-linux-${ARCH}.tar.xz"
+
+# Install libssl1.1
+RUN curl -fsSLO http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb \
+  && dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb \
+  && rm libssl1.1_1.1.1f-1ubuntu2_amd64.deb
```
aqrln, line 27: So it's not even in the repository, curious why so many users had it on their machines.

Also what I don't quite understand is that this package comes from Ubuntu 20.04, AFAICT, and yet there are no issues there.

aqrln (approved):

PR #5687, opened 2024-12-02
feat: drop Node.js 16

jkomyno:
This PR closes [ORM-396](https://linear.app/prisma-company/issue/ORM-396/investigate-and-fix-ecosystem-tests).

Only 18 tests are failing now, which is an improvement from the 214 previously failing tests.


docker-unsupported/amazonlinux-2-amd64-openssl-1.0.x/README.md
```diff
@@ -5,3 +5,10 @@
 - Arch: amd64 (x86_64)
 - OpenSSL location: `/lib64/libssl.so.10`
 - Binary target: `rhel-openssl-1.0.x`
+
+Running `prisma` -v will fail with:
+
+```sh
+node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by node)
+node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by node)
+```
```
aqrln, line 14: I think you might've actually built and run this as ARM64 and not AMD64 image. On x86_64, we target glibc 2.17, so you wouldn't get these errors. Our ARM binaries, however, do require a recent-ish glibc.

aqrln (approved):
I have one comment but let's merge this ASAP and iterate later

docker-unsupported/amazonlinux-2-amd64-openssl-1.0.x/README.md
jkomyno: I'll add a follow-up later

PR #5898, opened 2025-03-04
chore: use ubuntu 22.04

FGoessler:


.github/workflows/test.yaml
```diff
@@ -1716,13 +1716,13 @@ jobs:
           - debian-latest-arm64-openssl-3.0.x
           - opensuse-tumbleweed-amd64-openssl-3.0.x
           - redhat-ubi9-amd64-openssl-3.0.x
-          - ubuntu-20.04-amd64-openssl-1.1.x
+          - ubuntu-22.04-amd64-openssl-1.1.x
           - ubuntu-22.04-amd64-openssl-3.0.x
           - ubuntu-22.04-amd64-openssl-1.1.x
```
jacek-prisma: `ubuntu-22.04-amd64-openssl-1.1.x` is repeated twice

jacek-prisma: we might want to leave the old ubuntu test in place, I think it runs inside of docker anyway

jacek-prisma:
I think `binaries/pkg/test.sh` needs updating

jacek-prisma (approved):

PR #5900, opened 2025-03-04
fix: add an npmrc that disables side effect cache

jacek-prisma:
Fixes [platforms-serverless (netlify-github-with-nextjs-caching, library, ubuntu-22.04)](https://github.com/prisma/ecosystem-tests/actions/runs/13657259829/job/38179781400#logs)

### DIFF
--- a/platforms-serverless/netlify-github-with-nextjs-caching/.npmrc
+++ b/platforms-serverless/netlify-github-with-nextjs-caching/.npmrc
@@ -1,0 +1,1 @@
+side-effects-cache = false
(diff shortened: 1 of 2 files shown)


FGoessler:
Are those flags required now to make prisma work at all on netlify or why exactly do we need them? 🤔 

If yes and we have no other way we should ensure that this is at least documented because otherwise our users will have a very bad experience and nearly no chance of figuring this out o their own. 😬

jacek-prisma:
@FGoessler these flags seem to be necessary to have pnpm re-run the generate script, I agree it might be worth documenting somewhere. Basically the issue we're seeing is that if you don't include `prisma generate` in your `package.json` build script it won't get executed from a postinstall script and the prisma client won't exist when deployed.

jacek-prisma:
We decide that we'll delete this test because we want to get away from relying on install scripts

PR #5907, opened 2025-03-05
fix: get libsql to work on lambda

jacek-prisma:
Turso, when imported via the default `libsq/client` and built with esbuild doesn't work on lambdas. [There's an open issue for it](https://github.com/tursodatabase/libsql-client-ts/issues/112). There is a workaround though - importing `libsql/client/web` directly, but it doesn't work when other packages import libsql as well, and they do in our case (like our adapter). I've managed to get it to work through a hacky esbuild alias `--alias:'@libsql/client=@libsql/client/web'`.

Fixes [driver-adapters (turso-lambda-basic, library, ubuntu-22.04)](https://github.com/prisma/ecosystem-tests/actions/runs/13680152516/job/38250207569?pr=5907#logs)

### DIFF
--- a/driver-adapters/turso-lambda-basic/index.js
+++ b/driver-adapters/turso-lambda-basic/index.js
@@ -1,6 +1,6 @@
 // @ts-check
 const { Prisma, PrismaClient } = require('@prisma/client')
-const { createClient } = require('@libsql/client/web')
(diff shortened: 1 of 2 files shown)


jacek-prisma:
Fixed in prisma/prisma#26526

PR #5908, opened 2025-03-05
fix: move unsupported/debian-buster-amd64-openssl-1.1.x to supported

aqrln:
While the operating system used for the base image in this test uses OpenSSL 1.1 indeed, what this test actually checks is the scenario where OpenSSL is not installed at all. This would previously fail but now works with vendored OpenSSL.

Refs: https://linear.app/prisma-company/issue/ORM-770/fix-docker-tests


jacek-prisma:
I think you'll need to update `.github/workflows/test.yaml`

jacek-prisma (approved):

PR #5923, opened 2025-03-07
chore: create CI cron job to cleanup left over cloud resources

FGoessler:
This PR adds an updated scripted CI job to cleanup left over cloud resources in azure and gcloud.

It runs on a cron timer once a week but can also be triggered manually via the workflow UI. 

Why can there be left over resources in the first place you ask?
=> Mainly because if a test fails it can be helpful to have the cloud resources incl logs not be autodeleted. So in case of failures we keep them around. In success case we already delete them.


Nova-ZA:
Has these scripts been tested?

FGoessler:
> Has these scripts been tested?

Yes, I ran them even via CI prior to the cleanup rebase. :)

.github/scripts/wipe-azure.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
aqrln:
```suggestion
#!/usr/bin/env bash
```

`/bin/bash` is not portable, various systems have it in different locations. Only `/bin/sh` is guaranteed to be in `/bin` as required by POSIX, and `env` being in `/usr/bin` is a de-facto standard.

For example, on my Linux system bash is `/run/current-system/sw/bin/bash` and `/bin/bash` doesn't exist.

.github/scripts/wipe-azure.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Check if the argument is valid
+if [ -z "$FN_RESOURCE_GROUP" ]; then
+    echo "Usage: $0 <resource-group>"
+    echo "Usually <resource-group> is either prisma-e2e-windows or prisma-e2e-linux"
+    exit 1
+fi
+
+# Ensure the user is logged in
+if ! az account show > /dev/null 2>&1; then
```
aqrln: You can do this in modern bash (since 4.0, i.e. any version released in the last 16 years):

```suggestion
if ! az account show &> /dev/null then
```

The only reason not to use this is if you want to keep compatibility with the ancient bash 3 shipped with macOS by default (which Apple will never upgrade for license reasons, they don't like GPLv3) — I normally don't care about it, bash 3 is barely usable nowadays anyway (it doesn't even support `**` globs), and I'd assume everyone installs a usable bash via Homebrew at least at some point anyway. This is another reason why using `/usr/bin/env` in the shebang is important, you always want to use `/opt/homebrew/bin/bash` on macOS if it exists and not `/bin/bash`.

.github/scripts/wipe-azure.sh
aqrln: (non blocking here because we know `/bin/bash` exists on CI, but generally important in other cases)

.github/scripts/wipe-azure.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# Exit on error
+set -e
```
aqrln: maybe
```suggestion
set -euo pipefail
```
for more safety

`set -u` errors on undefined variables, `set -o pipefail` propagates the errors out of `|` pipelines instead of silently eating them.

.github/scripts/wipe-azure.sh
```diff
@@ -0,0 +1,54 @@
...
+    exit 1
+fi
+
+# Get the subscription ID
+SUBSCRIPTION_ID=$(az account show --query "id" -o tsv)
+echo "Using subscription: $SUBSCRIPTION_ID"
+
+# Get all Function Apps in the current subscription
+echo "Fetching all Function Apps in resource group $FN_RESOURCE_GROUP..."
+FUNCTION_APPS=$(az functionapp list --resource-group $FN_RESOURCE_GROUP --query "[].{name:name, resourceGroup:resourceGroup, appInsights:appInsights}" -o json)
```
aqrln: nit
```suggestion
FUNCTION_APPS=$(az functionapp list --resource-group "$FN_RESOURCE_GROUP" --query "[].{name:name, resourceGroup:resourceGroup, appInsights:appInsights}" -o json)
```

.github/scripts/wipe-gcloud.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
aqrln, line 1:
```suggestion
#!/usr/bin/env bash
```

.github/scripts/wipe-gcloud.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Exit immediately if a command exits with a non-zero status
+set -e
```
aqrln, line 4:
```suggestion
set -euo pipefail
```

aqrln (approved):
LGTM with a couple of nits for the second file as well

PR #5951, opened 2025-03-21
fix: update driver adapter creation

jacek-prisma:
Fixes a number of tests by updating how we create the adapters


viktor-ferenczi (approved):
Looks good. 

Cannot test it locally, relying on CI.

jkomyno (approved):

PR #5958, opened 2025-04-03
fix: modernize and fix Deno tests

aqrln:
- Switch from Deno 1.x to Deno 2.x.
- Switch from `prisma-client-js` to `prisma-client-ts` generator.
- Use `deno.json` instead of `package.json`.
- Disable running Deno with binary engine because there is an issue with ESM `binary.mjs` bundle, at least on Deno. We should take a look into it before GAing the new generator.

Closes: https://linear.app/prisma-company/issue/ORM-843/fix-deno-ecosystem-tests


FGoessler (approved):
This also switches the test to deno 2.x. Are we "breaking" deno 1.x functionality?

aqrln:
> This also switches the test to deno 2.x. Are we "breaking" deno 1.x functionality?

We never officially supported it (despite having the ecosystem tests) so I guess it's fine either way. I believe Deno 1.x should probably still work, the tests were mainly broken because they accidentally relied on the auto-installation functionality (which is going to be restored to fix other ecosystem tests), but it's more important for us to test Deno 2.x with ESM client rather than obsolete Deno 1.x with the old client (so this PR also contributes to https://linear.app/prisma-company/issue/ORM-683/e2e-tests-esm).

PR #5979, opened 2025-08-21
test: run DA tests with the query compiler

jacek-prisma:
Runs some of the driver adapter tests with the query compiler. I was hoping to get more of them working but I ended up excluding vercel and lambda tests for now. That's because they only work with the new generator. I've attempted changing the test to use the new generator dynamically, but it ended up being very complicated and not always working. If we want to have lambda/vercel coverage with the new generator, I think we should just introduce new tests.

Cool fact: the TIDB driver adapter works with the query compiler with some minor changes (and it's run within this test matrix).


jacek-prisma:
The 2 failing deno tests are not related and always fail

FGoessler (approved):

